Issue #39, opened 2022-02-21
avoid non working tutorials 


daviddoji:
I'd like to stress this point. If one want to build trust with potential users, the pieces you provide with the library should work "out of the box".
 
The only tutorial attached so far (new.ipynb) relies on some data that is not available with the library. There are a couple of options:
- Either the path from where the dataset could be downloaded must be specified at the beginning of the notebook
- The dataset is also provided with this library.

In any case, for the time being, I would suggest to hide the tutorial until the above point is clarified.

Any thoughts @steinnymir @zainsohail04 ?

steinnymir:
yes, you are perfectly right. This tutorial was put there as a placeholder I think.

I am removing them for the time being,

PR #40, opened 2022-02-21
remove placeholder tutorials

steinnymir:
As suggested, this removes the tutorials which show no real functionality.

These will be replaced in duetime with working examples.


steinnymir:
closes #39 by removing tutorials.

PR #43, opened 2022-02-24
Generalized histogram binning function

daviddoji:
I've made a generalization of the binning function. Now it can take any number of dimensions.
There are still a couple of details I'm not yet fully convinced:
- The name of the function: I personally do not like it but I kept the naming convention of the functions I replaced. Could we come up with something shorter or more descriptive?
- The way we have to pass the sample argument. Even if we want to calculate a 1D histogram, the sample has to be passed as a 2D array (np.array[[....],1]). I couldn't find a proper way to remove that and be compatible with Numba :(

Let me know what you think before merging.

P.S.: I've added also a couple of tests for that function.
P.S.2.: Sorry for the bad naming of the commit. I didn't squash them before pushing :(

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -32,9 +32,34 @@ python-versions = ">=3.6"
 name = "colorama"
 version = "0.4.4"
 description = "Cross-platform colored terminal text."
-category = "dev"
+category = "main"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+
+[[package]]
+name = "commonmark"
+version = "0.9.1"
+description = "Python parser for the CommonMark Markdown spec"
+category = "dev"
+optional = false
+python-versions = "*"
+
+[package.extras]
+test = ["flake8 (==3.7.8)", "hypothesis (==3.55.3)"]
+
+[[package]]
+name = "coverage"
+version = "6.3.2"
+description = "Code coverage measurement for Python"
+category = "dev"
+optional = false
+python-versions = ">=3.7"
+
+[package.dependencies]
+tomli = {version = "*", optional = true, markers = "extra == \"toml\""}
+
+[package.extras]
+toml = ["tomli"]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


steinnymir:
Thanks, this generalized function looks very nice! 
I agree with the naming, and maybe we could call this _hist_linear. 
We could then add a _hist_from_array, which instead of bins and ranges takes an array of values to be used as bins (for non-linear and non-uniform binnings).
If you come up with better names, go ahead! this is just a suggestion.

It makes sense that the input must be 2D, as it is in principle a table, and the 1D case is sort of a degenerate version of a table. So the requirement is also logically ok.

steinnymir (changes requested):
Maybe should have commented here... 
Anyway, you can change the names of the functions then its a go for me!

daviddoji:
> Thanks, this generalized function looks very nice! I agree with the naming, and maybe we could call this _hist_linear. We could then add a _hist_from_array, which instead of bins and ranges takes an array of values to be used as bins (for non-linear and non-uniform binnings). If you come up with better names, go ahead! this is just a suggestion.
> 
> It makes sense that the input must be 2D, as it is in principle a table, and the 1D case is sort of a degenerate version of a table. So the requirement is also logically ok.

If we want to make another function, such as the one you propose, I would name this one ```_hist_from_bins_ranges```. 
Another option would be, as it does basically the same as ```histogramdd``` from numpy but returning ints instead of floats, we could name it as ```_histogram_ints```

steinnymir:
This function does not actually work like histogramdd yet.
numba_histogramdd is close as it has a similar input and output.
However, also this one is still lacking the array input for binning as well as weighted binning, which is also on the todo list.

I agree the name `_hist_from_bins_ranges` is a good one, and the others could be `_hist_from_arrays`. 
The weighted versions could be `_hist_from_bins_ranges_weights` and `_hist_from_arrays_weights`.
All these are then used in the `numba_histogramdd `which we could simply call `histogramdd`, and add it as API `sed.histogramdd`.
What do you think?

Issue #32, opened 2022-02-08
Documentation


steinnymir:
Find a good method to automise the creation of the documentation page.

Also, we need to fix the commenting style to use when commenting code. Suggestions?

daviddoji:
AFAIK, documentation is created from .rst files and built automatically with sphinx

zainsohail04:
> AFAIK, documentation is created from .rst files and built automatically with sphinx

You're right. We're just thinking if there is a more automated way to document. Sphinx does have autodoc which I think is good to use. I was thinking if the type hints can be incorporated somehow and it seems sphinx extension Napoleon might be able to do it

PR #45, opened 2022-03-05
Documentation generation

zainsohail04:
#32 suggests to implement an automated way for documentation. So far, it is automated to the level of each module within the package. We need to manually add an rst file for each module that is added. I have automated the arg types so that typehints are used to automatically include that information, as well as the default values. One thing to keep in mind is that the return docstring doesn't support multi value output so it has to be wrapped within a tuple.

The matplotlib dependency is also within the pull request.


sed/binning.py
```diff
@@ -25,36 +24,36 @@ def _arraysum(array_a, array_b):
...
+    nBins: int | Sequence[int] = 100,
     hist_mode: str = "numba",
     jitterParams: dict = None,
     return_edges: bool = False,
-) -> np.ndarray:
+) -> np.ndarray | tuple[np.ndarray, list]:
     """Compute the n-dimensional histogram of a single dataframe partition.
 
     Args:
-        part (ddf.DataFrame): dataframe on which to perform the histogram.
```
daviddoji, line 40: why types were removed everywhere?

sed/binning.py
```diff
@@ -1,7 +1,6 @@
 # All functions in this file are adapted from https://github.com/mpes-kit/mpes
 from functools import reduce
 from typing import Sequence
-from typing import Tuple
```
daviddoji, line 4: if we stick to Python 3.8, this import is needed

pyproject.toml
```diff
@@ -17,6 +17,11 @@ xarray = "^0.21.1"
 numba = "^0.55.1"
 threadpoolctl = "^3.1.0"
 tqdm = "^4.62.3"
```
daviddoji, line 19: That's ok, but next time make another PR if two different issues are addressed, e.g. addind new dependency and creating documentation.
It's easier for review ;)

sed/binning.py
zainsohail04: As mentioned above 

> I have automated the arg types so that typehints are used to automatically include that information, as well as the default values.

There is no longer a need for including the types twice henceforth.

pyproject.toml
zainsohail04: Right. Most of the dependencies added are for documentation purposes. I saw that you included matplotlib in the other PR so I think it can be removed from here.

sed/binning.py
daviddoji: Although `pyenv` makes it really easy to choose any Python version, we should stick not to the bleeding-edge ones. 
I chose 3.8 because it is the one installed on Maxwell.

rettigl: Do you use 3.8 features that are not part of 3.7? I would stick to the lowest version that provides the features necessary for the code. The old packages run just fine in 3.7. I would not break this compatibility without very good reason

zainsohail04: > Do you use 3.8 features that are not part of 3.7? I would stick to the lowest version that provides the features necessary for the code. The old packages run just fine in 3.7. I would not break this compatibility without very good reason

Currently not in use but the importlib.metadata can end up being quite useful in many situations. For example to use in single source versioning throughout the package. 
I am not aware if any other features from 3.8 are or will be used. With 3.10, I already had some reason to have used it.

daviddoji: > Do you use 3.8 features that are not part of 3.7? I would stick to the lowest version that provides the features necessary for the code. The old packages run just fine in 3.7. I would not break this compatibility without very good reason

I chose 3.8 for two reasons:
- It's the default in Maxwell when you spawn a node
- Some dependencies (pandas, dask, ...) require, at least, python 3.8 in their last releases. As in some point, we will need to update it anyway, I decided to do it the sooner the better considering this library is still under heavy development

steinnymir: I also agree that python 3.8 is the better choice. 
Earlier versions will only cause us to need to update sooner than later.
Bleeding edge versions like 3.10 are not easy to handle on the user side, outweighting the advantages posed by the new features.

sed/binning.py
steinnymir: This will be a little unconfortable, but I guess its ok.

I usually autogenerate the docstring, and this includes the types by default. Maybe I can find some setting that changes that.

steinnymir: I thought this would be unconfortable when using autogenerated docstring strucutres, but it turns out there is a setting to use google-notypes (or whatever base style) which does the job!

steinnymir (approved):
All looks good

zainsohail04:
> All looks good to me, so once the conflict is resolved, feel free to merge from my side. @rettigl and @zainsohail04? your reviews are still open, do you agree with the changes proposed?

Since I am the author, I can not also be the reviewer of this. However, I agreed with @daviddoji and changed the version to 3.8 already.

daviddoji (approved):

daviddoji:
> > All looks good to me, so once the conflict is resolved, feel free to merge from my side. @rettigl and @zainsohail04? your reviews are still open, do you agree with the changes proposed?
> 
> Since I am the author, I can not also be the reviewer of this. However, I agreed with @daviddoji and changed the version to 3.8 already.

I think you'll need to checkout the `poetry.lock` and `pyproject.toml` files to resolve the conflicts

rettigl (approved):
I don't overlook all changes here, but sure, go ahead. One further comment: Does in addition to the poetry stuff the standard pip installation still work?

daviddoji:
> I don't overlook all changes here, but sure, go ahead. One further comment: Does in addition to the poetry stuff the standard pip installation still work?

Do you mean installing the dependencies from a `requirements.txt` files? That can easily be done, if people find it useful.
Do you mind @rettigl to open an issue so we keep track of this request?

PR #46, opened 2022-03-08
Setting up a virtual env with poetry

daviddoji:
This PR aims at making a reproducible working environment where all the current dependencies needed to run the library and the example notebooks are met.

Instructions about how to set it up were added to the README.


daviddoji:
BTW, merged PRs can be deleted

zainsohail04:
For users of Maxwell, I believe this setup procedure (README) is quite complex and doesn't allow for intuitive integration with other packages. This guide should serve really well for future developers trying to setup a development environment but I'd argue for scientific users, packaging to PyPi and giving instructions to just install that is a much simpler procedure for most. This also leads to jupyter and ipykernel not being a necessary dependencies. What do you think?

daviddoji:
> For users of Maxwell, I believe this setup procedure (README) is quite complex and doesn't allow for intuitive integration with other packages. This guide should serve really well for future developers trying to setup a development environment but I'd argue for scientific users, packaging to PyPi and giving instructions to just install that is a much simpler procedure for most. This also leads to jupyter and ipykernel not being a necessary dependencies. What do you think?

Regarding complexity of the setup, I agree that it's a bit more complex than a simple `pip install` but the benefits of poetry clearly overcomes that.

We can do a `requirements.txt`, so users can install dependencies of the library using `pip`.

I included jupyter and ipykernel because:
- There are/will be notebooks in the library
- In order to create a kernel with all dependencies used by the library
We can put them as `dev` dependencies

zainsohail04 (approved):
We can go ahead with this then.
Next, I can setup a github action to publish the next version to pypi and then we can update the README to accomodate that as well. 
Regarding putting those dependencies as dev, they can rather be put as optional jupyter dependencies (poetry allows for specifying different optional groups).

steinnymir (approved):
The poetry installation instructions are not nearly as complex as those we had for hextof-processor. I believe these to be very good for direct detailed usage of this package.

On the other hand, we should also provide, as suggested, an easier `pip install` method for those who only want to use the simple API, or for importing as dependency in other projects.

Also, I very much like adding the `dev` tag to install the jupyter part, as this might not be always required and adds a lot of dependencies.

daviddoji:
> The poetry installation instructions are not nearly as complex as those we had for hextof-processor. I believe these to be very good for direct detailed usage of this package.
> 
> On the other hand, we should also provide, as suggested, an easier `pip install` method for those who only want to use the simple API, or for importing as dependency in other projects.
> 
> Also, I very much like adding the `dev` tag to install the jupyter part, as this might not be always required and adds a lot of dependencies.

The creation of a `pip install sed` package, of course, can be (and should be) done. But I would wait a bit longer untill more functionalities are integrated in the library.

Regarding jupyter dependencies, I'm inclined towards leave them as they are now because in my workflow Jupyter is heavily used and I don't know another easy way to create a kernel with all the dependencies needed by the library. But, I'm also open to put them as `dev` dependencies due to the reasons given by @steinnymir

steinnymir:
I totally agree, I would also always install jupyter and its extensions, therefore we can have it as default.

PR #48, opened 2022-03-10
Set jupyter and ipykernel as optional dependencies

zainsohail04:
I decided to create a PR to clarify what I meant by optional dependency. Please check the README here as well to see if this workflow makes sense or not. I personally think it makes sense thinking long term when we put this on PyPi, where SED will just be one package in the user's workflow. Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,9 +20,13 @@ poetry shell
 poetry install
 ```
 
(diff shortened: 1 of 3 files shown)


daviddoji (approved):
As said in https://github.com/OpenCOMPES/sed/pull/46, that's fine for me.

LGTM!

Issue #26, opened 2022-02-08
Allow non-equidistant binning


steinnymir:
This is possible with the numpy binning approach (hextof-processor method), but needs a cleaner interface.

Need to look into the numba implementation to see if this is possible there too.

steinnymir:
Solved in #50

PR #50, opened 2022-03-11
Arbitrary binning axes

steinnymir:
Implementing binning with arbitrary axes, allowing for non-linear binnings.

It works for bin_partition, but still not for bin_dataframe.
The reason is I am trying to use nBins as either the number of bins in the old method, but also to pass it arrays directly.
This then drops the requirement of binRanges. I will try fix this....

However the numba function works as of now, even if I had to use a workaround for the creation of the target nd-array.
I always hit this problem where you cannot create tuples in Numba.

@daviddoji if you can have a look at this would be great! the tutorial notebook has tests which use this new function.

### DIFF
--- a/sed/binning.py
+++ b/sed/binning.py
@@ -419,8 +419,9 @@ def _hist_from_bin_range(
             j = (sample[t, i] - ranges[i, 0]) * delta[i]
             is_inside = is_inside and (0 <= j < bins[i])
             flatidx += int(j) * strides[i]
+            # don't check all axes if you already know you're out of the range
             if not is_inside:
-                break  # don't check all axes if you already know you're out of the range
+                break  
         if is_inside:
             Hflat[flatidx] += int(is_inside)
 
@@ -490,12 +491,12 @@ def _hist_from_bins(
         flatidx = 0
         for i in range(ndims):
             j = binsearch(bins[i], sample[t, i])
-            is_inside = is_inside and (
-                j >= 0
-            )  # binsearch returns -1 when the value is outside the bin range
+            # binsearch returns -1 when the value is outside the bin range
+            is_inside = is_inside and (j >= 0)  
             flatidx += int(j) * strides[i]
+            # don't check all axes if you already know you're out of the range
             if not is_inside:
-                break  # don't check all axes if you already know you're out of the range
+                break  
         if is_inside:
             Hflat[flatidx] += int(is_inside)
 
@@ -507,7 +508,7 @@ def numba_histogramdd(
     bins: Union[Sequence[Union[int, np.ndarray]], np.ndarray],
     ranges: Sequence = None,
 ) -> Tuple[np.ndarray, np.ndarray]:
-    """Wrapper for the Number pre-compiled binning functions.
+    """ Multidimensional histogramming function, powered by Numba.
 
     Behaves in total much like numpy.histogramdd. Returns uint32 arrays.
     This was chosen because it has a significant performance improvement over
@@ -554,7 +555,6 @@ def numba_histogramdd(
             method = "int"
         elif isinstance(bins[0], np.ndarray):
             method = "array"
-
     except AttributeError:
         # bins is a single integer
         bins = D * [bins]
@@ -575,7 +575,6 @@ def numba_histogramdd(
         return hist, bins
 
     elif method == "int":
-
         # normalize the range argument
         if ranges is None:
             ranges = (None,) * D
--- a/sed/binning.py
+++ b/sed/binning.py
@@ -19,10 +19,7 @@ N_CPU = psutil.cpu_count()
 
 
 def _arraysum(array_a, array_b):
-    """
-    Calculate the sum of two arrays.
-    """
-
+    """ Calculate the sum of two arrays."""
     return array_a + array_b
 
 
@@ -432,7 +429,7 @@ def _hist_from_bin_range(
 
 @numba.jit(nogil=True, parallel=False, nopython=True)
 def binsearch(bins: np.ndarray, val: float) -> int:
-    """Bisection index search function.
+    """ Bisection index search function.
 
     Finds the index of the bin with the highest value below val, i.e. the left edge.
     returns -1 when the value is outside the bin range.
@@ -488,7 +485,6 @@ def _hist_from_bins(
 
     for i in range(ndims):
         strides[i] = H.strides[i] // H.itemsize
-
     for t in range(sample.shape[0]):
         is_inside = True
         flatidx = 0
--- a/sed/binning.py
+++ b/sed/binning.py
@@ -19,19 +19,25 @@ N_CPU = psutil.cpu_count()
 
 
 def _arraysum(array_a, array_b):
-    """ Calculate the sum of two arrays."""
+    """Calculate the sum of two arrays."""
     return array_a + array_b
 
 
 def _simplify_binning_arguments(
     bins: Union[
-        int, dict, tuple, List[int], List[np.ndarray], List[tuple],
+        int,
+        dict,
+        tuple,
+        List[int],
+        List[np.ndarray],
+        List[tuple],
     ] = 100,
     axes: Union[str, Sequence[str]] = None,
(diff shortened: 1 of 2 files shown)


sed/binning.py
```diff
@@ -199,8 +199,8 @@ def bin_dataframe(
                         dfPartition,
                         None,
                         binAxes,
-                        binRanges,
                         nBins,
+                        binRanges,
                         hist_mode,
                     ),
                 )
```
daviddoji, line 206: Is there any good reason to keep `legacy` and `lean` modes, apart from compatibility?

sed/binning.py
```diff
@@ -355,7 +355,85 @@ def _hist_from_bin_ranges(
             j = (sample[t, i] - ranges[i, 0]) * delta[i]
             is_inside = is_inside and (0 <= j < bins[i])
             flatidx += int(j) * strides[i]
+            if not is_inside:
```
daviddoji, line 358: I'd put comments before the statements. In this case, before the `if not is_inside`

sed/binning.py
```diff
@@ -355,7 +355,85 @@ def _hist_from_bin_ranges(
...
+                break  # don't check all axes if you already know you're out of the range
+        if is_inside:
+            Hflat[flatidx] += int(is_inside)
+
+    return H
+
+
+@numba.jit(nogil=True, parallel=False, nopython=True)
+def binsearch(bins: np.ndarray, val: float) -> int:
```
daviddoji, line 367: is this similar to the already implemented `bisect_left` from the standard library (https://docs.python.org/3/library/bisect.html)?

sed/binning.py
```diff
@@ -355,7 +355,85 @@ def _hist_from_bin_ranges(
...
+    ndims = len(bins)
+    if sample.shape[1] != ndims:
+        raise ValueError(
+            "The dimension of bins is not equal to the dimension of the sample x",
+        )
+    H = np.zeros(shape, np.uint32)
+    Hflat = H.ravel()
+
+    strides = np.zeros(ndims, np.int64)
+    # print(H.shape,strides.shape,H.strides.shape,H.itemsize)
```
daviddoji, line 421: Do not forget to remove commented lines used for debugging :)

sed/binning.py
```diff
@@ -364,9 +442,9 @@ def _hist_from_bin_ranges(
 
 def numba_histogramdd(
     sample: np.array,
-    bins: Sequence,
-    ranges: Sequence,
-) -> Tuple[np.array, np.array]:
+    bins: Union[Sequence[Union[int, np.ndarray]], np.ndarray],
+    ranges: Sequence = None,
+) -> Tuple[np.ndarray, np.ndarray]:
     """Wrapper for the Number pre-compiled binning functions.
```
daviddoji, line 448: I would skip the reason why the function was created. The user do not care :)
Regarding the overflow, I would use an assert for that

sed/binning.py
```diff
@@ -405,37 +483,56 @@ def numba_histogramdd(
...
-                " sample x.",
-            )
-    except TypeError:
-        # bins is an integer
+        Db = len(bins)
+        if isinstance(bins[0], int):
+            method = "int"
+        elif isinstance(bins[0], np.ndarray):
+            method = "array"
+
```
daviddoji, line 491: remove empty line

sed/binning.py
```diff
@@ -405,37 +483,56 @@ def numba_histogramdd(
...
-        M = len(bins)
-        if M != D:
-            raise ValueError(
-                "The dimension of bins must be equal to the dimension of the "
-                " sample x.",
-            )
-    except TypeError:
-        # bins is an integer
+        Db = len(bins)
+        if isinstance(bins[0], int):
```
daviddoji, line 487: would be simpler if `bins` are always either `ints` or `arrays`??

daviddoji (commented):
Apart from the comments, I'm also missing tests for the functions

sed/binning.py
```diff
@@ -28,8 +28,8 @@ def bin_partition(
     part: Union[dask.dataframe.core.DataFrame, pd.DataFrame],
     binDict: dict = None,
```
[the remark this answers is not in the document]
steinnymir: removed!

sed/binning.py
steinnymir: They were there mostly to test which one performs best. For the time being I would leave them

sed/binning.py
steinnymir: good point! I will look into the function descriptions again

steinnymir:
I changed the function argument names to better follow the numpy naming in similar functions.
Also added multiple ways to define the bins:
- an integer as before
- a tuple
- a pre-defined numpy array
all the above can be passed directly or as a list of them.

Finally, one can also pass a dictionary with keys the names of the columns and values being any of the above.
For now, the same type needs  to be used for all the dimensions on which we are binning, i.e. no mixing between the three types above.

I started making a test function, but have not "tested" it yet. I will complete the pull request once all functionalities have a test function. This is the right approach right?

sed/binning.py
steinnymir: Not sure what you mean...

sed/binning.py
steinnymir: yep, looks exactly the same! however:

TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Untyped global name 'bisect_left': Cannot determine Numba type of <class 'builtin_function_or_method'>

File "..\sed\binning.py", line 495:
def _hist_from_bins(
    <source elided>
            # j = binsearch(bins[i], sample[t, i])
            j = bisect_left(bins[i], sample[t, i])
            ^

Numba doesn't know bisect_left, so I had to do it manually.

steinnymir:
The test functions I made don't work. I guess I need to learn more about testing.... @daviddoji could you have a look and tell me what I am doing wrong there? 😓 

I tested it in the tutorial, which changed accordingly.

Meanwhile I added also the jitter functionality, so now the binning features should be complete!
This would close #26

sed/binning.py
daviddoji: yeah, numba is a bit picky with that. You can try this one from numpy:
https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html

According to numba's docs, it's supported:
https://numba.readthedocs.io/en/stable/reference/numpysupported.html?highlight=numpy%20

Sorry for being harsh on this, but my point on this is to reuse well-tested functions instead of creating them by ourselves :)

sed/binning.py
daviddoji: why is it necessary to support either bins or arrays?
Would not be easier to support only arrays?

sed/binning.py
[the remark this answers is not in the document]
daviddoji: My apologies @steinnymir. I meant to use snake case, as it is the standard in the Python community

sed/binning.py
steinnymir: we have two binning methods now, implemented with numba. One based on defining the number of bins (therefore int) and is much faster. The other takes a pre-made array and is slower, but allows non-linear binning.

Therefore the two different input methods. Also, this mimics what numpy.histogramdd does, where you can pass an int or an array.

rettigl:
I played around a bit with this. The version that uses the bin edges array as input does not produce the same output as the version with start and enpoints and number of bins. Specifically, there seems to be a problem with the last bin, which contained 0 in my test in the version using the bin edges.
On the plus side, I did not observe any performance differences between the two approaches (however, still with equidistant bins).
Also, just passing a list of the axes on which to apply jittering does not do anything at the moment, seems this is not implemented. I will have a look at the version with the dictionary.

daviddoji:
is this PR ready for merging? @steinnymir

steinnymir:
> The version that uses the bin edges array as input does not produce the same output as the version with start and enpoints and number of bins

@rettigl I also observed this once, but I cannot reproduce it any more. Looking at real data, binned with both methods and with the hextof library, all give the same exact result.

Therefore for now I merge this pull request, and if this problem reappears, we can add an issue for it.

PR #57, opened 2022-03-24
xarray to h5 saving/loading functions 

rettigl:
with recursive parsing of metadata dictionary
There certainly is a lot to improve, and this does not work for all data types in the metadata dictionary (some are converted to strings). We can generate multiple export/import functions. This is what we use currently to further convert to Nexus (we have a working pipeline for this).

### DIFF
--- a/sed/io.py
+++ b/sed/io.py
@@ -85,47 +85,6 @@ def recursive_parse_metadata(
     return dictionary
 
 
-def res_to_xarray(
-    res: np.array,
-    bin_names: list,
-    bin_axes: list,
-    metadata: dict = None,
-) -> xr.DataArray:
-    """creates a BinnedArray (xarray subclass) out of the given np.array
-
-    Parameters:
-        res: nd array of binned data
-        bin_names: list of names of the binned axes
-        bin_axes: list of np.arrays with the values of the axes
-        metadata: Dictionary containing the metadata
-
-    Returns:
-        xres: BinnedArray
-            an xarray DataArray container with binned data, axis, and all available
-            metadata
-    """
-    dims = bin_names
-    coords = {}
-    for name, vals in zip(bin_names, bin_axes):
-        coords[name] = vals
-
-    xres = xr.DataArray(res, dims=dims, coords=coords)
-
-    for name in bin_names:
-        try:
-            xres[name].attrs["unit"] = DEFAULT_UNITS[name]
-        except KeyError:
-            pass
-
-    xres.attrs["units"] = "counts"
-    xres.attrs["long_name"] = "photoelectron counts"
-
-    if metadata is not None:
-        xres.attrs["metadata"] = metadata
-
-    return xres
-
-
 def xarray_to_h5(data: xr.DataArray, faddr: str, mode: str = "w"):
     """Save xarray formatted data to hdf5
 
@@ -145,7 +104,12 @@ def xarray_to_h5(data: xr.DataArray, faddr: str, mode: str = "w"):
         print(f"saving data to {faddr}")
 
         # Saving data, make a single dataset
-        h5File.create_dataset("binned/BinnedData", data=data.data)
+        dataset = h5File.create_dataset("binned/BinnedData", data=data.data)
+        try:
+            dataset.attrs["units"] = data.attrs["units"]
+            dataset.attrs["long_name"] = data.attrs["long_name"]
+        except KeyError:
+            pass
 
         # Saving axes
         axesGroup = h5File.create_group("axes")
@@ -156,6 +120,10 @@ def xarray_to_h5(data: xr.DataArray, faddr: str, mode: str = "w"):
                 data=data.coords[binName],
             )
             axis.attrs["name"] = binName
+            try:
+                axis.attrs["unit"] = data.coords[binName].attrs["unit"]
+            except KeyError:
+                pass
             axesNumber += 1
 
         if "metadata" in data.attrs and isinstance(
@@ -170,7 +138,7 @@ def xarray_to_h5(data: xr.DataArray, faddr: str, mode: str = "w"):
 
 
 def h5_to_xarray(faddr: str, mode: str = "r") -> xr.DataArray:
-    """Rear xarray data from formatted hdf5 file
+    """Read xarray data from formatted hdf5 file
 
     Args:
         faddr: complete file name (including path)
@@ -189,21 +157,42 @@ def h5_to_xarray(faddr: str, mode: str = "r") -> xr.DataArray:
             )
 
         # Reading the axes
-        axes = []
-        bin_names = []
+        binAxes = []
+        binNames = []
 
         try:
             for axis in h5_file["axes"]:
-                axes.append(h5_file["axes"][axis])
-                bin_names.append(h5_file["axes"][axis].attrs["name"])
+                binAxes.append(h5_file["axes"][axis])
+                binNames.append(h5_file["axes"][axis].attrs["name"])
         except KeyError:
             raise Exception("Wrong Data Format, the axes were not found.")
 
         # load metadata
+        metadata = None
         if "metadata" in h5_file:
             metadata = recursive_parse_metadata(h5_file["metadata"])
-            xarray = res_to_xarray(data, bin_names, axes, metadata)
 
-        else:
-            xarray = res_to_xarray(data, bin_names, axes)
+        coords = {}
+        for name, vals in zip(binNames, binAxes):
+            coords[name] = vals
+
+        xarray = xr.DataArray(data, dims=binNames, coords=coords)
+
+        try:
+            for name in binNames:
+                xarray[name].attrs["unit"] = h5_file["axes"][axis].attrs[
+                    "unit"
+                ]
+            xarray.attrs["units"] = h5_file["binned"]["BinnedData"].attrs[
+                "units"
+            ]
+            xarray.attrs["long_name"] = h5_file["binned"]["BinnedData"].attrs[
+                "long_name"
+            ]
+        except KeyError:
+            pass
+
+        if metadata is not None:
+            xarray.attrs["metadata"] = metadata
+
         return xarray
--- a/sed/io.py
+++ b/sed/io.py
@@ -6,8 +6,7 @@ import xarray as xr
 
 
 def recursive_write_metadata(h5group: h5py.Group, node: dict):
-    """
-    Recurces through a python dictionary and writes it into an hdf5 file.
+    """Recurces through a python dictionary and writes it into an hdf5 file.
 
     Args:
         h5group: hdf5 group element where to store the current dict node to.
@@ -26,7 +25,7 @@ def recursive_write_metadata(h5group: h5py.Group, node: dict):
                 h5group.create_dataset(key, data=item)
             except TypeError:
                 h5group.create_dataset(key, data=str(item))
-                print("saved " + key + " as string")
+                print(f"Saved {key} as string.")
         elif isinstance(item, dict):
             print(key)
             group = h5group.create_group(key)
@@ -34,16 +33,17 @@ def recursive_write_metadata(h5group: h5py.Group, node: dict):
         else:
             try:
                 h5group.create_dataset(key, data=str(item))
-                print("saved " + key + " as string")
+                print(f"Saved {key} as string.")
             except BaseException:
-                raise ValueError("Cannot save %s type" % type(item))
+                raise Exception(
+                    f"Unknown error occured, cannot save {item} of type {type(item)}.",
+                )
 
 
 def recursive_parse_metadata(
     node: Union[h5py.Group, h5py.Dataset],
 ) -> dict:
-    """
-    Recurces through an hdf5 file, and parse it into a dictionary.
+    """Recurces through an hdf5 file, and parse it into a dictionary.
 
     Args:
         node: hdf5 group or dataset to parse into dictionary.
@@ -113,9 +113,9 @@ def xarray_to_h5(data: xr.DataArray, faddr: str, mode: str = "w"):
             data.attrs["metadata"],
             dict,
         ):
-            meta_group = h5File.create_group("metadata")
+            metaGroup = h5File.create_group("metadata")
 
-            recursive_write_metadata(meta_group, data.attrs["metadata"])
+            recursive_write_metadata(metaGroup, data.attrs["metadata"])
 
     print("Saving complete!")
 
--- a/sed/io.py
+++ b/sed/io.py
@@ -1,10 +1,44 @@
-import h5py
-import xarray as xr
-import numpy as np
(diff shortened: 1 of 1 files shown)


sed/io.py
```diff
@@ -1 +1,163 @@
...
+    'dldTimeBinSize': 'ns',
+    'delay': 'ps',
+    'timeStamp': 's',
+    'energy': 'eV',
+    'kx': '1/A',
+    'ky': '1/A'}
+
+def res_to_xarray(res, bin_names, bin_axes, metadata=None):
+    """ creates a BinnedArray (xarray subclass) out of the given np.array
+    Parameters:
```
daviddoji, line 60: For consistency, IMHO it's better to use a single style for docstrings

sed/io.py
```diff
@@ -1 +1,163 @@
...
+        mode (str): hdf5 read/write mode
+    Returns:
+    """
+    with h5py.File(faddr, mode) as h5File:
+
+        print(f'saving data to {faddr}')
+
+        # Saving data
+
+        ff = h5File.create_group('binned')
```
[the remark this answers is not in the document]
daviddoji: Actually, you don't need to create a variable if used only once. h5py allows you to do something like `h5File.create_dataset("binned/BinnedData", ...)

sed/io.py
```diff
@@ -1 +1,163 @@
...
+import h5py
+import xarray as xr
+import numpy as np
+
+def recursive_write_metadata(h5group, node):
+    for key, item in node.items():
+        if isinstance(item, (np.ndarray, np.int64, np.float64, str, bytes, int, float, list)):
+            try:
+                h5group.create_dataset(key, data=item)
+            except TypeError:
```
daviddoji, line 10: this is checked twice, here and in line 19

sed/io.py
```diff
@@ -1 +1,163 @@
...
+    Returns:
+        xarray (xarray.DataArray): output xarra data
+    """
+    with h5py.File(faddr, mode) as h5_file:
+        # Reading data array
+        try:
+            data = h5_file['binned']['BinnedData']
+        except KeyError:
+            print("Wrong Data Format, data not found")
+            raise
```
daviddoji, line 141: why to re-raise the exception when you can print the error directly in the print statement?

daviddoji (commented):
Hi @rettigl ! I took the liberty to suggest some improvements :)

sed/io.py
rettigl: They are different cases. Line 10 handles the case that one of the known data types cannot be saved as number. Line 19 tries to save everything unknown as string, and reports an error if also this fails. But maybe the logic could be changed.

sed/io.py
```diff
@@ -1 +1,163 @@
...
+            dictionary = dictionary.item()
+            if isinstance(dictionary, (bytes, bytearray)):
+                dictionary = dictionary.decode()
+        except ValueError:
+            pass
+
+    return dictionary
+
+
+DEFAULT_UNITS = {
```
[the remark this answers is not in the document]
rettigl: > Sure

sed/io.py
rettigl: I'll try to keep the style of the binning.py. Is this somehow auto-formatted?

steinnymir: I used the `google-notypes` style (google, which does not add the types in the comments), and auto-formatting exists for visual studio code as the extension autoDocstring, or somewhere in the settings of pycharm.

steinnymir:
This looks great! Thanks @rettigl. I did not have much time to look at it yet, but I hope to find time to test it next week!

sed/io.py
rettigl: You mean in the raise Exception("") statement? I want to create an error with program abortion, but provide a meaningful error message. I understand that is the way to do it?

sed/io.py
```diff
@@ -1 +1,163 @@
-# TODO: populate
+import h5py
+import xarray as xr
+import numpy as np
```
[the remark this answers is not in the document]
rettigl: I tried to take this into account, not sure it did what it's supposed to do. I'm not so comfortable with lots of auto-thingies tweaking around my code...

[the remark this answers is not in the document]
daviddoji: Great. Let me know if you need a help or want a quick demo.
Linters just do formatting, they do not change code logic

sed/io.py
daviddoji: > I'll try to keep the style of the binning.py. Is this somehow auto-formatted?

What I meant with that is the following:
- Start the docstring with capitalize letters.
- Use "camel_case" for naming variables, functions, ...
- In the `binning` module, "Args" is used instead of "Parameters"

I understand that these are "nitpicking" issues, but it's better to addressed them now that there is little code written :)

sed/io.py
daviddoji: Yep. You use the `try...except` block but then, when the `try` fails due to a KeyError, you print the message and then "re-raise" the exception.
Why not using :
```
except KeyError as e:
    print(f"Wrong data format. Expected ... but got {e}")
```

sed/io.py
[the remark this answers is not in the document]
rettigl: Maybe you can clarify the workflow a bit for me, to me it seems difficult to comprehend and document all the steps the code goes through with these pre-commit hooks.
When I add a changed file and commit, the pre-commit hooks lint (i.e. change) the file, but at this stage I cannot commit the unchanged version, becaue the pre-commit hook fails. So then I need to re-add the then-changed files in order to be able to commit again, but the changes the pre-commit hooks do, are not reflected in github. Is this the intended behavior? I would rather like to go step by step, commit my initial version, lint it, commit the linted version.

sed/io.py
rettigl: Fair points, I will go through once more.

sed/io.py
rettigl: But the problem with this is, that it consumes the Exception, and continues code execution, no? This condition should be a fatal abort condition, as the data format does not fit for loading the file. So I want to abort with raising a fitting exception. I understood this is how I do that.

sed/io.py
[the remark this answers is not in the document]
steinnymir: Onestly I am also a bit confused on how the pre-commit hooks actually work. I have failed in making them run automatically when I commit. I am running them manually every time now, which is not exactly practical.

Maybe a comprehensive short explaination of how to handle this in the wiki here would be very beneficial! @daviddoji would you be willing to make something like that?

sed/io.py
steinnymir: I agree that you still need the `raise` inside the `except` as you want to break code execution.
I whould however print the error message you want to show inside the error. For example:
```
try:
    data = h5_file['binned']['BinnedData']
except KeyError as e:
    raise KeyError(f"Wrong data format. The HDF5 file selected does not match the expected structure. \n Original error: {e}")
```

sed/io.py
```diff
@@ -75,89 +113,97 @@ def res_to_xarray(res, bin_names, bin_axes, metadata=None):
...
     if metadata is not None:
-        xres.attrs['metadata'] = metadata
+        xres.attrs["metadata"] = metadata
 
     return xres
 
-def xarray_to_h5(data, faddr, mode='w'):
-    """ Save xarray formatted data to hdf5
+
+def xarray_to_h5(data: xr.DataArray, faddr: str, mode: str = "w"):
```
daviddoji, line 71: Do we foresee any other mode than writing? If not, I would simplify the API and remove the mode from here

sed/io.py
```diff
@@ -39,32 +85,24 @@ def recursive_parse_metadata(node):
...
-
-def res_to_xarray(res, bin_names, bin_axes, metadata=None):
-    """ creates a BinnedArray (xarray subclass) out of the given np.array
+def res_to_xarray(
+    res: np.array,
+    bin_names: list,
+    bin_axes: list,
+    metadata: dict = None,
+) -> xr.DataArray:
+    """creates a BinnedArray (xarray subclass) out of the given np.array
```
daviddoji: Capitalize the first letter of docstring

sed/io.py
```diff
@@ -1,10 +1,44 @@
...
+    "timeStamp": "s",
+    "energy": "eV",
+    "kx": "1/A",
+    "ky": "1/A",
+}
+
+
+def recursive_write_metadata(h5group: h5py.Group, node: dict):
+    """
+    Recurces through a python dictionary and writes it into an hdf5 file.
```
daviddoji, line 10: Docstring starts in previous line, the one with the quotes :)

sed/io.py
```diff
@@ -18,10 +52,22 @@ def recursive_write_metadata(h5group, node):
...
-                raise ValueError('Cannot save %s type'%type(item))
+            except BaseException:
+                raise ValueError("Cannot save %s type" % type(item))
+
+
+def recursive_parse_metadata(
+    node: Union[h5py.Group, h5py.Dataset],
+) -> dict:
+    """
+    Recurces through an hdf5 file, and parse it into a dictionary.
```
daviddoji, line 46: Same here

sed/io.py
```diff
@@ -18,10 +52,22 @@ def recursive_write_metadata(h5group, node):
             try:
                 h5group.create_dataset(key, data=str(item))
                 print("saved " + key + " as string")
-            except:
-                raise ValueError('Cannot save %s type'%type(item))
+            except BaseException:
+                raise ValueError("Cannot save %s type" % type(item))
```
daviddoji, line 39: I would use an f-string here.

sed/io.py
```diff
@@ -18,10 +52,22 @@ def recursive_write_metadata(h5group, node):
             try:
                 h5group.create_dataset(key, data=str(item))
                 print("saved " + key + " as string")
-            except:
-                raise ValueError('Cannot save %s type'%type(item))
+            except BaseException:
```
daviddoji, line 38: If there would be a BaseException, why raising a ValueError?

sed/io.py
```diff
@@ -75,89 +113,97 @@ def res_to_xarray(res, bin_names, bin_axes, metadata=None):
 
     for name in bin_names:
         try:
-            xres[name].attrs['unit'] = DEFAULT_UNITS[name]
+            xres[name].attrs["unit"] = DEFAULT_UNITS[name]
         except KeyError:
             pass
```
daviddoji: if we don't handle the error, why are we using a `try...except` block?

sed/io.py
```diff
@@ -75,89 +113,97 @@ def res_to_xarray(res, bin_names, bin_axes, metadata=None):
...
         faddr (str): complete file name (including path)
         mode (str): hdf5 read/write mode
+
+    Raises:
+        Warning: subfunction warns if elements have been converted into strings for
+        saving.
+
     Returns:
     """
     with h5py.File(faddr, mode) as h5File:
```
daviddoji, line 86: Remove empty line

sed/io.py
```diff
@@ -75,89 +113,97 @@ def res_to_xarray(res, bin_names, bin_axes, metadata=None):
...
-
-        # make a single dataset
-        ff.create_dataset('BinnedData', data=data.data)
+        # Saving data, make a single dataset
+        h5File.create_dataset("binned/BinnedData", data=data.data)
 
         # Saving axes
-        aa = h5File.create_group("axes")
-        ax_n = 0
+        axesGroup = h5File.create_group("axes")
```
daviddoji, line 98: camel_case is prefered for naming variables

sed/io.py
```diff
@@ -75,89 +113,97 @@ def res_to_xarray(res, bin_names, bin_axes, metadata=None):
...
-            print("Wrong Data Format, axes not found")
-            raise
+            raise Exception("Wrong Data Format, the axes were not found.")
 
         # load metadata
-        if 'metadata' in h5_file:
-            metadata = recursive_parse_metadata(h5_file['metadata'])
+        if "metadata" in h5_file:
+            metadata = recursive_parse_metadata(h5_file["metadata"])
             xarray = res_to_xarray(data, bin_names, axes, metadata)
```
daviddoji, line 157: remove empty line

sed/io.py
rettigl: I think, "w" overwrites files per default. If you want e.g. to append, you could change this with this option, that was the idea...

sed/io.py
rettigl: So what are the linters good for, if they don't fix this ;)

sed/io.py
rettigl: Fair point, I replaced this now with a generic exception, as this is anyways an unkown error.

sed/io.py
rettigl: This part is anyways gone now, but there are similar places: The idea is to silently skip over missing items in the dictionary. How would you do this better? Check with if clauses for the keys first?

sed/io.py
rettigl: Variables in binning.py are mostly using camelCase. function names are using snake_case. Should we stick to this, then?

sed/io.py
daviddoji: Yes, the "w" mode creates files from scratch. 
In my opinion, this should be only option. Giving too many choices to users might be not a good idea, because it increases the chance of breaking things and the complexity of the tests needed for those options.

sed/io.py
daviddoji: Linters only do "cosmetic" fixes. They are still not smart enough :)

sed/io.py
daviddoji: Yep, and "if not ..." should be ok

sed/io.py
daviddoji: I apologized by this confussion. snake_case is the prefered way of writing variables. CamelCase is used "only" in class names (https://pythonguides.com/python-naming-conventions/)

daviddoji (commented):
Thanks @rettigl

sed/io.py
rettigl: Sure, for this one cannot provide test for any possible case, but similarly not for any possible content of an xarray. I find it useful as an optional parameter, that typically anyways is omitted...

sed/io.py
rettigl: I read this that both are fine for variable names. If you want to change this now, we need to change it everywhere, also in binning.py, etc. I am not convinced that this is necessary. Maybe something to discuss in a meeting soon

daviddoji: You're absolutely right and I'll create a PR.
My point with this, as discussed with @steinnymir, is that for an external user of the library, if she/he sees inconsistent use of snake/camel case, docstrings with different styles, extensive tests, linting applied randomly, etc., it tells a bit about the care that the developers put into the library.

steinnymir (approved):
The h5 save/load functions look good. We should consider adding tests for them, but we can add an issue for that.

Issue #49, opened 2022-03-10
Add requirements.txt with minimum requirements necessary.


rettigl:
I cannot install the package with poetry:

```
Installing dependencies from lock file
Warning: The lock file is not up to date with the latest changes in pyproject.toml. You may be getting outdated dependencies. Run update to update them.

  SolverProblemError

  Because sed depends on ipykernel (^6.9.1) which doesn't match any versions, version solving failed.

  at /mnt/pcshare/users/Laurenz/AreaB/sed/.pyenv/lib/python3.8/site-packages/poetry/puzzle/solver.py:241 in _solve
      237│             packages = result.packages
      238│         except OverrideNeeded as e:
      239│             return self.solve_in_compatibility_mode(e.overrides, use_latest=use_latest)
      240│         except SolveFailure as e:
    → 241│             raise SolverProblemError(e)
      242│
      243│         results = dict(
      244│             depth_first_search(
      245│                 PackageNode(self._package, packages), aggregate_package_nodes

```
Also doing a pip -E install results in some dependency problems, it appears because dask is pinned to a specific version:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
distributed 2022.2.1 requires dask==2022.02.1, but you have dask 2021.12.0 which is incompatible.
```

 Why is this the case? I would remove any specific version pinning, if not absolutely required.

rettigl:
Okay, I understand now. 
`dask=^2021.12.0`
does not allow to jumpt to any 2022 version. In this case where they advance by year, we have to use >=.

zainsohail04:
Could you try once more, the lock file didn't merge properly last time.

daviddoji:
Anyway, I think it also makes sense to create a `requirements.txt` with pinned version.

steinnymir:
did anyone test this before closing it? I am having a really hard time to get the requirements installed together with the requirements of an other package (in this case MPES).

The requirements.txt is forcing conflicting versions, and therefore fails.
The poetry installation is still very counterintuitive to me, I cannot seem to find a way to make it work as intended.

I will continue trying, but this issue should be closed only once a well tested environment installation procedure has been identified and described.

Also, a conda install .yaml would be really helpful.

zainsohail04:
> did anyone test this before closing it? I am having a really hard time to get the requirements installed together with the requirements of an other package (in this case MPES).
> 
[...]

PR #58, opened 2022-03-27
Setup requirements

steinnymir:
Adding the `env.yml` file to create conda environments. 
This was created by hand based on the requirements.txt file, but worked when tested.

This conda approach is, in my opinion, a simpler and more user-friendly approach, which will make the package more usable for non-expert users before we provide a packaged version.

Also, I updated the readme with the corresponding instructions.

Please let me know what you think, and if you have suggestions on how to make this conda/pip/poetry more intuitive and flexible. 

This pull request addresses issue #49

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,26 @@ Single Event Data Frame Processor: Backend to handle photoelectron resolved data
 
 # Installation
 
+## Conda approach
+
+Clone this repository and cd to its root folder.
+Create a new environment by typing:
+```
+conda env create -f env.yml
+```
+This should install all the requirements to run `sed` in your environment.
+To activate your environment:
+```
+conda activate sed_conda
+```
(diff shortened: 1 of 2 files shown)


zainsohail04:
Seems like a good idea. Fixed versions is definitely good for development but I am not sure if it leads to conflicts when installing sed in another environment.
Currently, I tested by creating the environment as defined in the README and also installed hextof-processor package in the same environment without any issues.

zainsohail04 (approved):
Adding Python 3.8 to the environment might be a good idea.

steinnymir:
> 

As long as you create the environment from this file, fixed versions should not create problems. Installing requirements for other pakcages on top will then overwrite what is defined here eventually, and if conflicts arise, it is good to discover them soon.

Issue #62, opened 2022-04-06
Add data export to .tiff format  for use in ImageJ


steinnymir:
**Is your feature request related to a problem? Please describe.**
It is necessary to add  the tiff file as an output format.
This ensures to not break the workflow used by many users so far, which would hardly accept using our processing methods if they are not able to perform the analysis as they used to.

**Describe the solution you'd like**
define a function to save as .tiff

rettigl:
the mpes package has code for Tiff exporting. probably this can be reused:
```
   elif form == 'tiff': # Save as tiff stack

        try:
            import tifffile as ti
            ti.imsave(save_addr, data=dct[bdn].astype(dtyp))
        except ImportError:
            raise ImportError('tifffile package is not installed locally!')
```
This does not save the axes, which however I think is not supported by tiff.

steinnymir:
I also wrote a function to save to .tiff in hextof-processor, which ensures compatibility for 4D data with ImageJ. We can try both and pick the most versatile version, or make a combined version.

PR #65, opened 2022-04-06
Export_as_tiff

steinnymir:
Functions to export as tiff.
Still untested, once tested I will make this a proepr pull request.


steinnymir:
This closes #62

steinnymir:
I tested this and works as intended. I would make test functions, but I am unsure how that works with input-output of files...

rettigl:
An example of such a test function can be found here:
https://github.com/nomad-coe/nomad-parser-nexus/blob/master/tests/tools/dataconverter/test_writer.py

we could generate random data of varying dimensionality, write them and re-read them, and check that they are identical.

sed/io.py
```diff
@@ -179,3 +181,120 @@ def h5_to_xarray(faddr: str, mode: str = "r") -> xr.DataArray:
...
+        elif len(axis_name_list) == 1:
+            dims_order.append(*axis_name_list)
+        else:
+            dims_to_add[key] = 1
+            dims_order.append(key)
+
+    xres = data.expand_dims(dims_to_add)
+    xres = xres.transpose(*dims_order)
+    if ".tif" not in faddr:
+        faddr += ".tif"
```
rettigl, line 264: I would also allow ".tiff"

rettigl: Does this work if faddr is a Path?

rettigl: Also, maybe move this into the wrapper function.

steinnymir: Actually it does not work when faddr is Path, and all because of that if... I think I'll just remove it as it is there to ensure the file name has the right extension. There is probably a better way to do it with Pathlib.

daviddoji (commented):
Please, always write tests :)

sed/io.py
steinnymir: Removed this confusing if statement, replacing it with faddr.with_suffix('.tiff'), which ensures the file will have the .tiff extension.

sed/io.py
```diff
@@ -179,3 +181,120 @@ def h5_to_xarray(faddr: str, mode: str = "r") -> xr.DataArray:
             xarray.attrs["metadata"] = metadata
 
         return xarray
+
+
+def array_to_tiff(array: np.ndarray, faddr: Union[Path, str]) -> None:
```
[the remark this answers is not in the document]
steinnymir: The logic of how it saves is different. The xarray_to_tiff requires axes information and reorders the data according to the format ImageJ wants. 
Anyway I followed your suggestion and added handling of np.arrays in the xarray_to_tiff function.
Also, I renamed all functions dropping the "xarray" part.

steinnymir:
Added the test functions! 🎉 
but of course.. they don't  work. I get an error:
image

@daviddoji can you help me with this? it seems to "not find the test results"... happens for both test_diagnostics and test_io.

daviddoji:
> Added the test functions! tada but of course.. they don't work. I get an error: image
> 
> @daviddoji can you help me with this? it seems to "not find the test results"... happens for both test_diagnostics and test_io. The first iteration works, but all the others fail.

Sure :) 
What I see in `test_diagnostics.py` is that you're using `random.shuffle` in the wrong way. It's a method that works modifying the sequences "in place" (https://docs.python.org/3/library/random.html#random.shuffle). Therefore, it returns "None" when you use it that way.
If you want to shuffle, try this:
```
axes_names = ["x", "y", "t", "e"]
random.shuffle(axes_names)
```

Second issue is the following. As you want to ensure the lengths of "shape" and "dims" that you pass to the function "simulate_binned_data" are equal, until you pass a list of four elements, the first 3 tests fails.

image

I'll continue on Monday ;)

sed/io.py
```diff
@@ -1,12 +1,16 @@
+from msilib import sequence
```
rettigl, line 1: My installation does not know msilib out of the box. Include corresponding requirements.

rettigl: [msilib](https://docs.python.org/3/library/msilib.html#module-msilib) — Read and write Microsoft Installer files
???
Why msi files? Please explain.
Also: Deprecated since version 3.11: The [msilib](https://docs.python.org/3/library/msilib.html#module-msilib) module is deprecated (see [PEP 594](https://www.python.org/dev/peps/pep-0594) for details).

rettigl: Why not typing.Sequence?

sed/io.py
```diff
@@ -120,7 +124,7 @@ def xarray_to_h5(data: xr.DataArray, faddr: str, mode: str = "w"):
     print("Saving complete!")
 
 
-def h5_to_xarray(faddr: str, mode: str = "r") -> xr.DataArray:
+def load_h5(faddr: str, mode: str = "r") -> xr.DataArray:
```
rettigl, line 127: Modify the tutorial accordingly

sed/io.py
steinnymir: This is some autocompletition error. was intended to be typing.Sequence

sed/io.py
steinnymir: Which tutorial? Tutorial 1 disappeared for some reason.. and there is no call to this in tutorial 2. If you have a copy of tutorial 1 can you re-add it?

sed/io.py
daviddoji: are you using the `pre-commit` hooks? Because that kind of mistakes are avoided by either `isort` or `black`

steinnymir: yes, I am, but it didn't catch this one as it was being used. There was a function which expected a `sequence` type instead of a `Sequence` type. 
It was a sneaky bug which passed through our pre-commit guards!

sed/io.py
```diff
@@ -179,3 +182,143 @@ def h5_to_xarray(faddr: str, mode: str = "r") -> xr.DataArray:
...
+        out = data.expand_dims(dims_to_add)
+        out = out.transpose(*dims_order).values
+    else:
+        raise TypeError(f"Cannot handle data of type {data.type}")
+
+    faddr = Path(faddr).with_suffix(".tiff")
+
+    tifffile.imwrite(faddr, out.astype(np.float32), imagej=True)
+    # clean up the temporary axes names
+    for ax in _imagej_axes_order:
```
rettigl, line 279: In our case, this removes also the "X" and "Y" axes...

sed/diagnostics.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+import xarray as xr
+
+
+def simulate_binned_data(shape: tuple, dims: list):
```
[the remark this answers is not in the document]
rettigl: I tried to move this a bit forward. The tiff saving from np.ndarrays was not working, neither the tests for it. I fixed this, and did some linting on it. The binning is still broken in this branch.

PR #68, opened 2022-04-11
Make numba_histogramdd work as np.histogramdd

### LINKED ISSUE #67: test_from_bins_equals_from_bin_range() failing from time to time
**Describe the bug**
The test function test_from_bins_equals_from_bin_range() fails from time to time with an assetion error:
```
./sed/tests/test_binning.py::test_from_bins_equals_from_bin_range Failed: [undefined]AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 2 / 87210 (0.00229%)
Max absolute difference: 4294967295
Max relative difference: 4.2949673e+09
 x: array([[[0, 0, 0, ..., 0, 0, 0],
        [0, 0, 0, ..., 0, 0, 0],
        [0, 0, 0, ..., 0, 0, 0],...
 y: array([[[0, 0, 0, ..., 0, 0, 0],
        [0, 0, 0, ..., 0, 0, 0],
        [0, 0, 0, ..., 0, 0, 0],...
def test_from_bins_equals_from_bin_range():
        H1 = _hist_from_bin_range(sample3d, bins3d, ranges3d)
        H2 = _hist_from_bins(sample3d, arrays3d, tuple(b.size for b in arrays3d))
>       np.testing.assert_allclose(H1, H2)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 2 / 87210 (0.00229%)
E       Max absolute difference: 4294967295
E       Max relative difference: 4.2949673e+09
E        x: array([[[0, 0, 0, ..., 0, 0, 0],
E               [0, 0, 0, ..., 0, 0, 0],
E               [0, 0, 0, ..., 0, 0, 0],...
E        y: array([[[0, 0, 0, ..., 0, 0, 0],
E               [0, 0, 0, ..., 0, 0, 0],
E               [0, 0, 0, ..., 0, 0, 0],...

sed/tests/test_binning.py:47: AssertionError
```

The number looks like an overflow error. Is this related to the different histogram behavior reported in #66?
**To Reproduce**
Execute the test mutliple times.

**Expected behavior**
The test should run through always

**Screenshots**
n/a

**Additional context**
We should adopt the test function to restrict the

steinnymir:
This pull request fixes the inconsistant behaviour of the function `_hist_from_bins`. The aim is to mimic the behaviour of `np.histogramdd` in 'numba_histogramdd'.

I believe this to be the best solution as it is easier to implement for non-linear bins. Converting a list of center points in a non-linear array to the ranges of the bins defined around them can lead to unexpected behaviour, different from the users expectation. 
By explicitly requiring the bins edges and not the bins centers leaves more space to the user to define peculiar binnings.

Furthermore, I added test functions, which ensure the behaviour of `numba_histogramdd` to behave as `np.histogramdd`.
The only remaining exception is the fact we do not (yet) support weights and density as np does.

If agreed upon, this closes #67 and the long discussion about the inconsistent binning behaviour


sed/binning.py
```diff
@@ -545,6 +555,7 @@ def _hist_from_bins(
         flatidx = 0
         for i in range(ndims):
             j = binsearch(bins[i], sample[t, i])
+            # j = np.searchsorted(bins[i],sample[t, i])
```
rettigl, line 558: remove comment

sed/binning.py
```diff
@@ -624,14 +636,16 @@ def numba_histogramdd(
...
             tuple(bins),
-            tuple(b.size for b in bins),
+            tuple(b.size - 1 for b in bins),
         )
         return hist, bins
 
     elif method == "int":
         # normalize the range argument
         if ranges is None:
             ranges = (None,) * D
```
rettigl, line 646: What good does this do? For which case is this intended?

rettigl (commented):
In general looks good to me. However a few things:
1. np.histdd includes as far as I know the rightmost edge in the last bin, your binsearch excludes it. This is probably not found by the sparse dataset of the tests. Can be fixed in inserting another test in line 508 and changing the second condition in line 509
2. Line 289: We need to decide what we want to return as axis coordinates. Using the edges will not work anymore. Do we want to return the center of the bins here? How does xarray actually handle plotting of 2D arrays with non-equidistant axes?

rettigl:
You can enforce a failed test by setting e.g. 
`sample1d = np.array([[x] for x in np.linspace(-2,2,1001)])`

rettigl:
I added a fix for this, please have a look.

sed/binning.py
steinnymir: Actually I cannot recall, I think it still remains from the MPES implementation. But it should never happen that ranges is None in the "int" method, so I will remove it

steinnymir:
> I added a fix for this, please have a look.

This indeed does what you were mentioning, and adds values on the right edge to the last bin.
However, even though this is the numpy default behaviour, do we want this? does it not make the last bin slightly bigger than all the others, leading to possible artifacts in analysis? what do you think?

sed/binning.py
steinnymir: done

sed/binning.py
steinnymir: actually, even though it might be redundant, I will add an error here in case this happens, as it should not.

rettigl:
> However, even though this is the numpy default behaviour, do we want this? does it not make the last bin slightly bigger than all the others, leading to possible artifacts in analysis? what do you think?

Well, it's only an infinitisimal increase, as it only adds the edge of "size" 0 itself. As the goal is to make the two functions equal in behavior, I would go for this. Otherwise, the tests might randomly fail, which is not what we want.

steinnymir:
Ok, I agree! I would wait for an opinion from the others and eventually merge.

sed/binning.py
```diff
@@ -624,14 +637,20 @@ def numba_histogramdd(
...
 
     elif method == "int":
         # normalize the range argument
         if ranges is None:
-            ranges = (None,) * D
+            raise ValueError(
+                "must define a value for ranges when bins is"
+                " the number of bins",
+            )
+        #     ranges = (None,) * D
```
daviddoji, line 651: please leave all commented lines out before commiting :)

daviddoji (commented):
In general LGTM!

A couple of details for this/future PRs:
- Always use code conventions for naming (snake_case for functions/arguments, camelCase for classes, ...)
- Do not rush in committing a PR if it's not ready, i.e., do not have print statements used for debugging, all unintended comments have been cleared out, linters have been passed, ...
- Do not make huge functions. If a function is long enough (more than 50-60-ish lines), it's better to break it down in smaller functions because:
  + Debugging will be faster
  + You can test corner cases easier
- Even if you copy the functionality from somewhere else, do not copy/paste directly. Code it yourself and think if everything makes sense.

rettigl:
> 2\. Do we want to return the center of the bins here? How does xarray actually handle plotting of 2D arrays with non-equidistant axes?

Xarray interprets axis nodes apparently as the bin centers, so we should calculate and return these. I will also crosscheck that this is the same convention in nexus, but I believe so.

steinnymir:
> Xarray interprets axis nodes apparently as the bin centers, so we should calculate and return these.

I thought the same, but we should probably not by default return the center values of the bins. The function `numba_histogramdd`, just as the original `numpy `version, returns the bins and the edges. 
When constructing the binned array the simplest solution would be as you suggest to return the center values of the bins. However we should leave the option of representing the axes as ranges (`xarray`allows this, see results of `xr.groupby`, similar to `pandas`). It might help in some special cases.

We could then add a function which computes the center bins for all axes and have it easy accessible. It could also be the same function that does this when generating the xarray in the first place.

tests/test_binning.py
```diff
@@ -1,44 +1,72 @@
...
     ids=lambda x: f"bins:{len(x)}",
 )
-def test_hist_Nd_error_is_raised(_samples, _bins):
+def test_histdd_error_is_raised(_samples, _bins):
     with pytest.raises(ValueError):
         if _samples.shape[1] == len(_bins):
             pytest.skip("Not of interest")
-        _hist_from_bin_ranges(_samples, _bins, ranges1d)
+
+        _hist_from_bin_range(_samples, _bins, ranges[0])
```
rettigl, line 40: This test fails for me, as long as numba @jit is turned on. I have no clue why. The old test function, which uses essentially the same input function does work also with numba, with no change to the code. Absolutely mysterious to me...

rettigl: That is indeed the reason. However, I don't know how to fix it using the current setup of the tests. I seem not to be able to convert this into a np.array and still keep the tests running. It always throws the error, no matter how many np.asarrays I put and where.

sed/binning.py
```diff
@@ -36,6 +35,7 @@ def _simplify_binning_arguments(
     axes: Union[str, Sequence[str]] = None,
     ranges: Sequence[Tuple[float, float]] = None,
 ) -> tuple:
+    """TODO needs docstring"""
```
rettigl, line 38: @steinnymir I don't undestand what this function is supposed to do, how it works, and why it is called both in bin_dataframe and bin_partition. Therefore I don't know where to reasonably place the extension from bin centers to bin edges. Please provide a docstring and explanation, and tests for all cases.

sed/binning.py
```diff
@@ -95,6 +96,45 @@ def _simplify_binning_arguments(
...
+    """Converts a list of bin edges into corresponding bin centers
+
+    Args:
+        bin_edges: 1d array of bin edges
+
+    Returns:
+        bin_centers: 1d array of bin centers
+    """
+
+    bin_centers = bin_edges[:-1] + np.diff(bin_edges) / 2
```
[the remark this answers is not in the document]
rettigl: The purpose is to use these functions to convert input bin centers to bin edges for the binning. This is nontrivial when the bin centers are non-uniform, which is the only reason to use input bins as a list.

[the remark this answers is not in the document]
daviddoji: My point was that a function should encapsulate a "complex" operation. IMHO, a sum is not complex enough to deserve a function by itself because, at the end of the day, a function call adds a bit of overhead.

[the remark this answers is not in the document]
steinnymir: In my opinion, these functions are very useful for inexperienced users, which just want to "convert the bins" without thinking about how to mathematically approach the problem, nor learn about all `numpy` functions.

We need to keep in mind that the main task of this package is to be user friendly. I don't believe that the overheads of a function call are ever going to impact the performance to a significant level, especially for functions you call once.

sed/binning.py
steinnymir: This allows to mimic the input arguments of numpy.histogramdd to the bin_dataframe and bin_partition functions.

I externalized this in a function to remove the overhead of running through this at each parallel iteration of bin_partition. I'll write the docstring for this, and tests as they might be very important here.

sed/binning.py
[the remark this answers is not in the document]
daviddoji: That function will be called from another "function".
Neither the user nor you will have to call it directly ever, so I don't see the need for it as it's just doing "a simple operation". That was point.

To be honest, it's not something I'd write but I don't have a strong opinion about keeping it, if all others seems to be happy about it.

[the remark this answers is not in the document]
steinnymir: I see your point, and could in general agree. However in this case, this function will be useful also outside this function which calls it, and having it accessible to the users is going to be helpful.

sed/binning.py
rettigl: I understand the general purpose, but not the exact function. Also, I am wondering why you call it both in bin_dataframe and bin_partition. Would it not be cleaner to restrict the possible inputs to bin_partition to those of np.histogram, and only call it in bin_dataframe?

sed/binning.py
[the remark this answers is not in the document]
daviddoji: And just to reinforced my point, the numbers in the following benchmark is crystal clear:
image

sed/binning.py
rettigl: I don't see the use case for using bin_partition standalone outside a test case, where one can use restricted input definition.

sed/binning.py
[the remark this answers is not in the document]
rettigl: You are comparing pears an oranges. The two functions don't compute the same if the input is not equidistant!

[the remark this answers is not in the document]
daviddoji: @That's not true in this particular case. The input/output is exactly the same for both functions.

Could you give another input, so I could make my mind clearly into the problem? I can't help you with the code if I don't understand the problem ;)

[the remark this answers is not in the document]
rettigl: Ahh, not I see what you are doing. Okay, I can adopt that. It's more difficult to understand (for me), but I can add a respective comment line. Anyways, these are very small fish to fry, and not much worth the discussion. This is called once. Let's concentrate on adding functionality, or improving efficiency on the big scale (without changing functionality).

sed/binning.py
steinnymir: It is called in both as bin_partition could be used standalone. I did in fact use it when using pandas dataframes, and not dask ones.

I agree that this function looks horrible like this, and if there is a more elegant way to check and transform inputs, I'm happy to change this. This was the way I found so far to ensure the same signature (almost) for `np.histogramdd`, `bin_partition` and `bin_dataframe`...

rettigl: I see. The reason I argue to call it only once is that I was wondering where to extend input bin-lists from bin-centers into bin-eges as we discussed. A natural place seemed for me this function, but this only works if it only is called once. I don't know how to handle this and where to place this call, if bin_partition can be called standalone, and should have the same functionalitiy. This would make things much more complicated here.

rettigl:
I think the bugfix is done now, so I suggest to merge this one and do further improvements and code revisions in a new branch/pull request. @steinnymir please merge if you agree.

sed/binning.py
steinnymir: This is indeed called only once. In bin_dataframe, all calls to bin_partition are passing the flag skipTest, which makes bin_partition ignore this function, and assume the input is already correct.

rettigl: I see. Regarding the points for the representation in the xarray, I still think this is a separate issue that we should address in a separate branch. I realized this is also not correct the way we have done it so far for the case of num_bins and ranges, because the bins edges returned by np.histdd enclose the given range, whereas the bin centers we returned included the same range. We need to discuss how to handle this, and what the user expected behavior should be.

rettigl:
I created a new issue for the further fixes

sed/binning.py
daviddoji: Please correct me if I'm wrong. Up to now, this `bin_dataframe` function returns bins in the same way as np.histogramdd does. If that's the case, IMO that should be the default.
If for some reason, you guys think there should be another way of returning bins, we could think of passing a flag to the function to return this non-defaults

steinnymir: I agree this is still an open question. This PR needs to go though, so I will merge this and we can start new PRs from the issue you created and making other ones.

PR #69, opened 2022-04-15
Binning with jitter on machine coordinates

rettigl:
This pull request implements jittering on the intrinsic machine coordinates within the wrapper class by introducing a copy of the dataframe with jittering applied to the relevant coordinates. The idea would be to do this step in the beginning, and apply all following dataframe operations on both dataframes. Then, at the binning step, by choosing either dataframe, one can still choose between jitter or no jitter. Only downside is that this will be an all or nothing choice, depending on which axes one chose to be jitterd in the beginning.
The changes also include a first working version of binning in the wrapper class.

### DIFF
--- a/sed/core.py
+++ b/sed/core.py
@@ -1,19 +1,18 @@
-from typing import Any
 from pathlib import Path
-from typing import Union
-from typing import Tuple
-from typing import Sequence
+from typing import Any
 from typing import List
+from typing import Sequence
+from typing import Tuple
+from typing import Union
 
 import numpy as np
 import pandas as pd
 import psutil
 import xarray as xr
-import dask
 
-from .metadata import MetaHandler
-from .dfops import apply_jitter
 from .binning import bin_dataframe
+from .dfops import apply_jitter
+from .metadata import MetaHandler
 
 N_CPU = psutil.cpu_count()
 
@@ -21,14 +20,19 @@ N_CPU = psutil.cpu_count()
 class SedProcessor:
     """[summary]"""
 
-    def __init__(self, df: pd.DataFrame = None, metadata: dict = {}, config: Union[dict, Path, str] = {}):
+    def __init__(
+        self,
+        df: pd.DataFrame = None,
+        metadata: dict = {},
+        config: Union[dict, Path, str] = {},
+    ):
 
         # TODO: handle/load config dict/file
         self._config = config
         if not isinstance(self._config, dict):
             self._config = {}
-        if 'num_cores' not in self._config.keys():
-            self._config['num_cores'] = N_CPU-1
+        if "num_cores" not in self._config.keys():
+            self._config["num_cores"] = N_CPU - 1
 
         self._dataframe = df
         self._dataframe_jittered = None
@@ -94,32 +98,38 @@ class SedProcessor:
         self._dataframe = pd.DataFrame(data)
 
     def gen_jittered_df(self, cols: Sequence[str] = None) -> None:
-        """Generate a a second dataframe, where the selected dataframe columes have jitter applied
+        """Generate a a second dataframe, where the selected dataframe
+         columns have jitter applied
 
         Args:
-            cols: the colums onto which to apply jitter. If omitted, the comlums are taken from the config.
+            cols: the colums onto which to apply jitter. If omitted,
+            the comlums are taken from the config.
 
         Returns:
             None
         """
         if cols is None:
-            try: 
-                cols = self._config['jitter_cols']
+            try:
+                cols = self._config["jitter_cols"]
             except KeyError:
-                cols = self._dataframe.columns #jitter all columns
-            
-        self._dataframe_jittered = self._dataframe.map_partitions(apply_jitter, cols=cols, cols_jittered=cols)
+                cols = self._dataframe.columns  # jitter all columns
 
+        self._dataframe_jittered = self._dataframe.map_partitions(
+            apply_jitter,
+            cols=cols,
+            cols_jittered=cols,
+        )
 
     def compute(
         self,
         bins: Union[
-        int,
-        dict,
-        tuple,
-        List[int],
-        List[np.ndarray],
-        List[tuple],] = 100,
+            int,
+            dict,
+            tuple,
+            List[int],
+            List[np.ndarray],
+            List[tuple],
+        ] = 100,
         axes: Union[str, Sequence[str]] = None,
         ranges: Sequence[Tuple[float, float]] = None,
         histMode: str = "numba",
@@ -147,20 +157,36 @@ class SedProcessor:
         Returns:
             [description]
         """
-        assert self._dataframe is not None, 'dataframe needs to be loaded first!'
+        assert (
+            self._dataframe is not None
+        ), "dataframe needs to be loaded first!"
 
         if nCores is None:
-            nCores = self._config['num_cores']
+            nCores = self._config["num_cores"]
 
         df = self._dataframe
 
         if jittered:
-            assert self._dataframe_jittered is not None, 'jittered dataframe needs to be generated first, use SedProcessor.gen_jittered_df()!'
+            assert self._dataframe_jittered is not None, (
+                "jittered dataframe needs to be generated first, "
+                "use SedProcessor.gen_jittered_df(cols)!"
+            )
             df = self._dataframe_jittered
-        
-        self._binned = bin_dataframe(df=df, bins=bins, axes=axes, ranges=ranges, histMode=histMode, mode=mode, pbar=pbar, nCores=nCores, nThreadsPerWorker=nThreadsPerWorker, threadpoolAPI=threadpoolAPI, **kwds)
-        return self._binned
 
+        self._binned = bin_dataframe(
+            df=df,
+            bins=bins,
+            axes=axes,
+            ranges=ranges,
+            histMode=histMode,
+            mode=mode,
+            pbar=pbar,
+            nCores=nCores,
+            nThreadsPerWorker=nThreadsPerWorker,
+            threadpoolAPI=threadpoolAPI,
+            **kwds,
+        )
+        return self._binned
 
     def add_dimension(self, name, range):
         if name in self._coordinates:
--- a/sed/core.py
+++ b/sed/core.py
@@ -31,8 +31,18 @@ class SedProcessor:
         self._config = config
         if not isinstance(self._config, dict):
             self._config = {}
+        if "hist_mode" not in self._config.keys():
+            self._config["hist_mode"] = "numba"
+        if "mode" not in self._config.keys():
+            self._config["mode"] = "fast"
+        if "pbar" not in self._config.keys():
+            self._config["pbar"] = True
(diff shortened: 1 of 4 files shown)


sed/core.py
```diff
@@ -76,17 +97,48 @@ def load(self, data: pd.DataFrame) -> None:
         """
         self._dataframe = pd.DataFrame(data)
 
+    def gen_jittered_df(self, cols: Sequence[str] = None) -> None:
```
steinnymir, line 100: Is there a strong reason for creating a second dataframe, or could we not add these as columns in the same dataframe and then at binning select the columns with '_jittered' in the name?

Also, it might be that when creating new columns with, for example, calibrations, based on one columns, and later, but before computation jittering the original column in place, would induce the jitter to be computed before the computation.
This would allow us to apply the jitter function optionally just before binning.

sed/core.py
```diff
@@ -76,17 +97,48 @@ def load(self, data: pd.DataFrame) -> None:
...
+        ] = 100,
+        axes: Union[str, Sequence[str]] = None,
+        ranges: Sequence[Tuple[float, float]] = None,
+        histMode: str = "numba",
+        mode: str = "fast",
         jittered: bool = True,
-        ncores: int = N_CPU,
-        pbenv: str = "classic",
+        pbar: bool = True,
+        nCores: int = None,
```
steinnymir, line 139: This is not a comment to this PR, but rather a general one. Do we want to stick to camelCase for function attributes and snake_case for variables? 
In any case, the parameter names used in the binning function should be the same. I'll set up an issue for this.

sed/core.py
```diff
@@ -98,15 +150,43 @@ def compute(
...
+        assert (
+            self._dataframe is not None
+        ), "dataframe needs to be loaded first!"
+
+        if nCores is None:
+            nCores = self._config["num_cores"]
+
+        df = self._dataframe
+
+        if jittered:
```
steinnymir, line 169: If we apply the jitter at last (see comment above) this would be where to do it I guess.

steinnymir (approved):
LGTM, just a few comments on the functionalities.

sed/core.py
```diff
@@ -12,13 +20,26 @@
 class SedProcessor:
     """[summary]"""
 
-    def __init__(self):
+    def __init__(
+        self,
+        df: pd.DataFrame = None,
+        metadata: dict = {},
+        config: Union[dict, Path, str] = {},
```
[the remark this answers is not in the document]
rettigl: This is just a dummy so far. My idea was to allow for ider providing a dictionary with the config values, or a hande/string to a file from which the config can be read. The config-handling should be implemented in a separate class, like the metadata, I suppose.

sed/core.py
```diff
@@ -12,13 +20,26 @@
 class SedProcessor:
     """[summary]"""
 
-    def __init__(self):
+    def __init__(
+        self,
+        df: pd.DataFrame = None,
```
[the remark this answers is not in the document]
rettigl: Somehow, when I had dask.dataframe.DataFrame there, it evaluated to Any, not sure why. With pd.DataFrame, it also works for dask dataframes... Not sure in any case how and where these typing rules are checked or enforced?

sed/core.py
rettigl: > Is there a strong reason for creating a second dataframe, or could we not add these as columns in the same dataframe and then at binning select the columns with '_jittered' in the name?

I had thought about this. This option has two disadvantages:
1. The binning of the "unjittered" case also has the computational overhead of the "jittered" case, as the jittering operations are part of the dataframe tasks (in practice, this limitation is somewhat irrellevant, as most of the time jittering will be used)
2. During binning, the names of the jittered and not-jittered dataframe columnes are different, so the user must be aware of this, and provide the corresponding name. Off course we could also rewrite the axes-names instead of the dataframe, but I found this solution cleaner. We could also drop it entirely, and only leave the jittering option as part of the dataframe operation queue, and remove it from the binning function.

> 
> Also, it might be that when creating new columns with, for example, calibrations, based on one columns, and later, but before computation jittering the original column in place, would induce the jitter to be computed before the computation. This would allow us to apply the jitter function optionally just before binning.

I also thought about this. This has the disadvantage, that this option then changes the dataframe in an irreversible manner. Afterwards, you cannot bin without jitter anymore. The binning function should not change the dataframe.

sed/core.py
rettigl: Right, this is just copyPaste

sed/core.py
rettigl: Possibly, but as commented above, I would not modify the df in the binning function.

sed/core.py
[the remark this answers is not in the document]
daviddoji: In Python, being a dinamically typed language, the interpreter does type checking only as code runs. 
Adding type hints has no runtime effect: they are only hints and are not enforced on their own.

They are just use by type checker, e.g., mypy, which gives the some advantages:
- They are helpful for documentation
- They help you to build and maintain a cleaner architecture of the library

But they are just recommendations :)

sed/core.py
daviddoji: That's why I suggest to always write the code by yourself, even if you copy it from somewhere else ;)

steinnymir: using all the cores freezes you computer. The best choice is N-1 cores. In some cases however, we have observed better performance using less cores, due to less overheads in the parallelization (I think..)

sed/core.py
steinnymir: I see your point, but still am not convinced:
1. the overhead of the jittered case applies only if you "use" the jittered column. If you don't, this is never computed as binning is computed on the subset of columns to be binned.
2. The user doesn't  need to specify the names. If we want to use the jittered ones (with jittered=True) then we only need to check for each binning axis called "name", if there is a column called "name_jittered" and pass that one instead.
3. dropping it entirely would be the cleanest, but as you mention, we don't want to modify the df in the process. However, we never overwrite the data from disk, so nothing stops us from reloading the data, or even keeping a pointer to the "clean" data and "resetting" a column if required.

rettigl: > * the overhead of the jittered case applies only if you "use" the jittered column. If you don't, this is never computed as binning is computed on the subset of columns to be binned.

Yes it is, I tried it out. You can do benchmarking yourself, or I can post code here if you want...
> 
> * The user doesn't  need to specify the names. If we want to use the jittered ones (with jittered=True) then we only need to check for each binning axis called "name", if there is a column called "name_jittered" and pass that one instead.

True, but I have the impression it would make the code a lot "messier", and in this case you also have to apply the jitter first. I find the solution with two dataframes cleaner. My idea was to apply all following df operations transparently on both dfs.
> 
> * dropping it entirely would be the cleanest, but as you mention, we don't want to modify the df in the process. However, we never overwrite the data from disk, so nothing stops us from reloading the data, or even keeping a pointer to the "clean" data and "resetting" a column if required.

Well, if we decide to drop the non-jittered dataframe at the point of applying jitter _within the dataframe operations chain_ then this is where alteration of the dataframe is expected, and this would be fine. But then we should remove the "jittered" option from the binning function. This would be the cleanest version, but for non-jittered binning one would need to start over. Actually, I prefer that one, now that I think more about it.

sed/core.py
rettigl: Every "worker"/Core needs to reserve a copy of the target array in memory. For large binning sets and machines with many cores, you quickly run out of memory, if you don't limit the number of workers (e.g. 10 GB array* 140 cores = 1.4 TB RAM...). Also, you increase the overhead for recombination.

daviddoji: > using all the cores freezes you computer. The best choice is N-1 cores. In some cases however, we have observed better performance using less cores, due to less overheads in the parallelization (I think..)

I assume from your words you expect the users are going to use this library locally. Only in that case, I would see a point for letting the number of cores as a parameter.

You're right regarding performance. There is an overhead due to parallelization but I thought the data we will be dealing with is big enough to compensate it.

daviddoji: > Every "worker"/Core needs to reserve a copy of the target array in memory. For large binning sets and machines with many cores, you quickly run out of memory, if you don't limit the number of workers (e.g. 10 GB array* 140 cores = 1.4 TB RAM...). Also, you increase the overhead for recombination.

Why you need to copy data for every worker? That's not optimal at all.
Shared memory is used especifically for that. But you should know in advance the size of your data, which we know at the time of reading it.
Have a look at the following library we developed at EuXFEL (https://github.com/European-XFEL/pasha)

rettigl: Here, the issue is not about distributing the _source_ data over mutliple worker threads, this is done via dask, where every worker gets one dataframe partition. But each worker generates counts distributed over the _whole_ volume of the output array. If you want to use one common data volume for this, you need heavy synchronization between threads, with insane overhead, therefore one volume per thread, and later recombination. This works, is fast and reliable, but explodes if you make the number of workers too large. 20 is a good upper limit.

rettigl: Some benchmarks I did once:
3D binning:
Binning 3D
Speedup 3D
4D binning:
Binning 4D
Speedup 4D

sed/core.py
rettigl: We could leave the jittering on the binned coordinates in the binning function as an (undocumented) **kwrds argument option also in the code, or would this be confusing?

sed/core.py
daviddoji: As said before, that's why the library mentioned before was built.
Dask is a black box where you have no control over the data/workers. I'm a bit biased against dask, because it was not easy to make it work in my use case. And it was not only me, my colleagues on the data analysis group do not use dask neither, because it's tricky to control it.
Simple parallezation and/or distributed computing is not so complicated in Python :)

steinnymir: Using shared memory would be really nice. We should have a look into this (you told me already, but I didn't yet find a chance to..).

rettigl: We can discuss this off course, but I don't see a strong reason to change our running system.
We use dataframes not only for parallelization, but also for the delayed JIT loading of the data (which don't fit into memory), and the transparent and flexible manipulation of dataframe columns before computation without modification of the code. I am not sure how to reproduce these features otherwise, and even if we could, what the benefit of doing so would be.
In any case, this pull request is not the right place for such a discussion.

steinnymir: First of all, as you correctly pointed out, this is the wrong place for such a discussion.
If we think we should look into this more, we should make an issue/discussion.

But as a quick response:
1. I am strongly against moving away from dask in this stage
2. We can test a binning aggregation method based on shared memory, but that would not remove the dask dependency as, like Laurenz pointed out, the delayed processing is fundamental for our workflow. It might be good to have both options maybe, but don't change a working system!

sed/core.py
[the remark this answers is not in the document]
rettigl: This I would leave for another PR, making the config into a proper class of its own...
I would merge this soon, if there are no further comments.

steinnymir:
I agree this looks good now, and as we discussed, we should merge this and make new smaller PR for the things we want to change.

PR #72, opened 2022-08-11
added github action for linting and testing

### LINKED ISSUE #64: Github Action for running tests
We should add a github action to automatically run tests after commit

rettigl:
closes #64

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,39 @@
+name: python
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        lfs: true
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v3
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        git lfs pull
+        python -m pip install --upgrade pip
+        pip install pycodestyle pylint mypy pytest types-PyYAML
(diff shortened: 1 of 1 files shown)


rettigl:
I defined all the linting steps as continue-on-error, such that the chain only fails when the tests fail. Still, the linting errors and warnings are available from the action run logs, and can be addressed one by one.

steinnymir (approved):
LGTM! A very nice improvement!

daviddoji:
I see the point of this PR in running an action after the commit for the test functionality.
But regarding the linting/formating,  isn't that what the .pre-commit-hooks.yaml is for?

rettigl:
In principle yes, but a) I am using other linters here, and b) not everybody strictly uses the pre-commit hooks and c) if we take care of all the linting before committing anything, we will not make any progress at all. I see a workflow much more efficient where you can also commit half-working code, code with # TODO comments, etc. All not possible with these linters...
Then we can take care of cleaning the (working) code afterwards. For me, it's definitely functionality before ethicists. Otherwise, I'm not willing to spend the time.

rettigl:
Also, the current pre-commit hooks apparently overlook a lot of problems. See output of the action run.

zainsohail04:
Generally, you don't have to manually correct for the pretty modifications. There's a package called black that does it automatically. I think it is already included in this repos pre commit config file.

rettigl:
Yes, this is included in the pre-commit hook. But pylint, mypy etc. comment on a lot more things, like type inconsistencies, missing docstrings, etc. These cannot be automatically fixed

PR #77, opened 2022-08-18
Fix tutorial nb

daviddoji:
This change fixes the tutorial 1 notebook.

@steinnymir please have a look and merge if you agree with the changes


.pre-commit-config.yaml
```diff
@@ -12,31 +12,30 @@ repos:
...
         args: [--py36-plus]
-- repo: https://gitlab.com/iamlikeme/nbhooks
-  rev: 1.0.0
-  hooks:
-  - id: nb-ensure-clean
-    args: [--meta, ExecuteTime]
+-   repo: https://github.com/kynan/nbstripout
+    rev: 0.6.0
+    hooks:
+    -   id: nbstripout
```
rettigl, line 41: Does this automatically clean the output? Do we want this? Do we not also want to allow to publish notebooks with the expected output present?

rettigl: IMHO the information of the old precommit hook that the notebook still contains output is sufficient. One can then decide to still keep it, or manually strip it. Just my 5 cent.

daviddoji: yes, it does. This is typical way of sharing a notebook, as not all renders work in the same way.

daviddoji:
There is a test failing looking for a function in the `diagnostics` module that is not there.

rettigl:
This is fixed in the other branch. Merge the other PR first, then merge main into this one. Then it should be fine.

daviddoji:
> This is fixed in the other branch. Merge the other PR first, then merge main into this one. Then it should be fine.

Could you guys please merge the branch with the fix into main? Main should be bug free!

steinnymir:
@daviddoji please fix the conflict and I'll merge it

daviddoji:
Usually, fixes are done only by the person who opens the PR. Otherwise, you duplicate the work ;)

PR #81, opened 2022-08-30
modularization refactoring

steinnymir:
Today we discussed with @zainsohail04 and @kutnyakhov that the idea of splitting all the functionalities of `sed` in different packages might not be viable at the moment.
This is simply due to the amount of overhead work needed to maintain these packages (multiple repeated packaging and environment tools, changes in one package requiring manually updating versions in others, etc..), which distracts from developing the important features we want to work on.

In this pull request I suggest a change to the structure of this package which makes it more modular. This way we will keep the same separation structure we discussed, but all in a single package, which we should rename `sed-processor`.

I suggest this as a starting point, and if in the future we have the feeling one of the sub-modules in this package should have a life of its own, we can easily detach it.

I do not want to impose this, and therefore I ask for the approval of everyone to make these changes permanent. If someone disagrees, with a valid argument, then I will delete this and proceed with the separated packages approach we had previously discussed.


rettigl:
We can do it that way, but then we deviate from the initial goal to have a package independent from MPES/Metis specific code. Also, we will need to discuss how to reflect different loaders etc. and lab-specificity here.

sed/calibration/energy.py
```diff
@@ -0,0 +1,112 @@
...
+    def kinetic_energy(self, **kwargs):
+        pars = {**self.parameters, **kwargs}
+
+        return self._new_df.map_partitions(
+            self.tof_to_energy,
+            pars["tof_axis_name"],
+            pars["time_offset"],
+            pars["energy_offset"],
+            pars["tof_path_length"],
+        )
```
rettigl, line 112: I'm not quite sure I understand how this is supposed to be used or working. Can you provide an example?

sed/calibration/energy.py
```diff
@@ -0,0 +1,112 @@
...
+            the values of tof_ns converted to energy
+        """
+        k = 0.5 * 1e18 * 9.10938e-31 / 1.602177e-19
+        return (
+            k * np.power(tof_length / (tof_ns - time_offset), 2.0)
+            - energy_offset
+        )
+
+    def kinetic_energy(self, **kwargs):
+        pars = {**self.parameters, **kwargs}
```
rettigl, line 104: Shouldn't you reference self._parameters from within the class?

sed/calibration/energy.py
```diff
@@ -0,0 +1,112 @@
...
+            self._calib_df["tof_step"] = self._calib_df[self.tof_axis_name]
+            self._calib_df["tof_ns"] = self._calib_df["tof_step"]
+
+    @property
+    def parameters(self):
+        return {**self._parameters, **self._cached_parameters}
+
+    def clear_cache(self):
+        """clear the cached datasets and collect garbage"""
+        del self._cached_parameters
```
rettigl, line 61: what are these cached_parameters good for? are the filled anywhere?

rettigl (commented):
See my comments in the discussion. It depends in which direction we want to move this package. If we stick to this plan, we will loose the cleanliness of "just a dataframe binning package", and again make it quite specific to the TOF/momentum microscope audience. This is fine with me, but we need to be clear on that.

rettigl:
In terms of the logic structure, we should also discuss then what the "core" or the package is. In my view, this would rather be the class that wraps everything together, and not the binning part. So the binning should go into a sub-package called .binning, and the current "core" be a part of that package

sed/calibration/energy.py
steinnymir: Actually it is not yet supposed to work. These were work-in-progress changes that sneaked in when I changed the structure.

I will reset this, and make a new PR with these changes once we decided on the structural changes.

sed/calibration/energy.py
steinnymir: yes!

sed/calibration/energy.py
steinnymir: As above, these are not implemented yet. The idea here is to handle long computations of averages and cache them for future use, when testing parameters.

steinnymir:
Thank you @rettigl for your comments.
I changed the structure to better show the modular structure. Each module is in its folder, with `core` being a module, just like binning, io etc...

In this structure, `core` contains the processor (wrapper that handles the pipeline) and metadata handling classes.

I propose this as an intermediate structure to be able to go on as smoothly as possible with the development of the tools we are trying to create now, i.e. the data processing pipeline for momentum microscopes. Extracting and making a single package of those modules which could have a purpose on their own, like binning, can be done later rather easily from this structure.

sed/calibration/energy.py
steinnymir: The calibration tools were removed and replaced with a placeholder.

zainsohail04:
I support these changes. The main reason being, it seems easy to maintain multiple packages but I remember how challenging it can be to even maintain one, where we have to publish to PyPi, set up documentation channels, make easy to use Readme and tutorials for users. All while making sure nothing breaks. Sure, these can be simplified by CI like github actions but in no way is it still not time consuming. Add in the fact that we would have a strong dependency of sub packages with the core package, so everytime we update any (like a bug or the likes), we have to make sure they all are on correct versions.

I think io and and config modules could be submodules of core, as they can not be seperated from it. Also, perhaps processor.py is a misleading name, at least for development purposes as it doesn't clearly tell what that module is doing. Unless I am mistaking what it's supposed to do.

PR #84, opened 2022-09-23
Create the Dataframe class

steinnymir:
Here I am developing the dataframe class. This will work as the backbone for handling the single-events, allowing to combine fast and slow data and adding functionalities such as column name aliases, physical units etc, as dicussed in #83

### DIFF
--- a/env.yml
+++ b/env.yml
@@ -26,10 +26,12 @@ dependencies:
 - pyarrow==7.0.0
 - pyparsing==3.0.7
 - python-dateutil==2.8.2
+- pytest==7.1.2
 - pytz==2021.3
 - pyyaml==6.0
 - setuptools-scm==6.4.2
 - six==1.16.0
+- tifffile==2020.10.1
 - threadpoolctl==3.1.0
 - tomli==2.0.1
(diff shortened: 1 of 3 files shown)


rettigl:
I find this idea interesting, but am wondering if this is not oversophisticated somewhat. You will have to implement every dataframe operation with these aliases (assignment, mapping, filtering, all the numpy mappings, etc.)...
I would suggest to rather implement this on the level of the config file.
E.g. introduce a config section [dataframe][tof_column] = "t"
and then e.g. the notation "@tof_column" for the axis, which is evaluated and replaced by our functions.
Also, would there be a strong disadvantage in duplicating DF columns on creation  of the DF?

steinnymir:
If only for the aliases, this is absolutely an overkill. However I am looking into how to combine dataframes with different metadata, and conserve such information in the dataframe.
The idea is to create new columns based on which value changes in the metadata.

This might also be an unnecessarily complex feature, but I do have use cases in which it would have been confortable to have.

steinnymir:
For the time being, this is oversophisticated.

PR #86, opened 2022-10-25
Linting

rettigl:
linting of the codebase to pass the test workflow


rettigl:
In this PR I tried solving all the linter issues. In particular for the typing issues, I had to rework some of the code. Please check the changes carefully, as I did not originally write much of the code myself, and might have changed some behavior unintentionally.

.github/workflows/pylint.yml
```diff
@@ -23,17 +23,17 @@ jobs:
...
-        python -m pycodestyle --ignore=E501,W503 sed tests 
+        python -m pycodestyle --ignore=E203,E501,W503 sed tests
     - name: pylint
-      continue-on-error: true
+      if: ${{ always() }}
       run: |
-        python -m pylint sed tests 
+        python -m pylint --good-names=i,j,k,ex,x,y,t,k,v,ax,df --disable=fixme sed tests
     - name: mypy
-      continue-on-error: true
```
zainsohail04, line 34: Why is continuing on error a good thing?

rettigl: The old chain was running every item irrespective of its success, but only the last item, the tests would make the entire chain fail.
Now, every item will make the chain fail, but still every item is being executed. If you have neither continue-on-error not if(alyways), the chain would stop on the first failing task, and one would e.g. not run the pytests, if there is a linting error.

sed/binning/numba_bin.py
```diff
@@ -177,68 +184,67 @@ def numba_histogramdd(
...
-        edges = D * [None]
-        nbin = np.empty(D, int)
+    # Create edge arrays
+    edges: List[Any] = []
+    nbin = np.empty(num_cols, int)
 
-        for i in range(D):
-            edges[i] = np.linspace(*ranges[i, :], bins[i] + 1)
+    for i in range(num_cols):
+        edges.append(np.linspace(ranges[i][0], ranges[i][0], bins[i] + 1))
```
steinnymir, line 241: How deos this work? linspace from and to the same value... The test passes though, and behaves as numpy. Am I not seeing something?

sed/core/processor.py
```diff
@@ -22,7 +26,7 @@
 class SedProcessor:
     """[summary]"""
 
-    def __init__(
+    def __init__(  # pylint: disable=dangerous-default-value
```
steinnymir, line 29: Instead of disabling the pylint warning, we could correct the initialization to: `metadata = None`, and `if metadata is None: metadata = {}`

sed/binning/numba_bin.py
steinnymir, line 241: How deos this work? linspace from and to the same value... The test passes though, and behaves as numpy. Am I not seeing something?

steinnymir (approved):
Thank you @rettigl for cleaning this up! It looks so much more readable this way! I left just 2 comments, where as one is checked in the tests and does not fail, I am not sure if it is my misunderstanding...

sed/core/processor.py
rettigl: Good suggestions, thanks.

PR #93, opened 2022-11-26
Workflow improvements

rettigl:
Further improvements to the mpes workflow


rettigl:
This PR contains now a variety of changes to various parts of the code and workflow. In particular, all dataframe manipulation functions are now separated, and default calibration and correction parameters can be loaded from the config file. 
I would like to merge this rather soon, and then make smaller PRs for further improvements.

sed/loader/base/loader.py
```diff
@@ -50,5 +50,38 @@ def read_dataframe(
...
         return None, None
 
+    @abstractmethod
+    def get_count_rate(
+        self,
+        fids: Sequence[int] = None,
+        **kwds,
+    ) -> Tuple[np.ndarray, np.ndarray]:
+        """
+        Create count rate data for the files specified in ``fids``.
```
zainsohail04, line 60: Perhaps a generailzed implementation can already be in the base loader and the further additions in the plugins?
I also see that in generic loader, it is just returing None.

zainsohail04: I can't quite comment on many of other additions but I checked them and looked fine.

I would really push the idea of some documentation, because the codebase ported from mpes is getting quite complex and would really benefit from just at least having the docstrings in another file. The documentation is already setup so should be easy to add further modules. I, otherwise, do plan to work on it #99 (after steinn's update). Let me know if you can't figure it out for now.

zainsohail04: An abstract class (in our case, the base loader) is indeed a template, but can also contain implementations. Even if it's implemented, it needs to be overloaded by the subclasses. But the benefit is it just reduces the same code occuring multiple times. The subclass can just call super().my_abstract_method() in it's implementation and that's a valid use case. 

But this comes down to personal preference then. I prefer when the abstract classes already contain a general implementation, and the subclasses just overriding/adding to those implementations.

steinnymir:
We discussed this already a while ago, but I feel the need to remark this again. A pull request of this size is impossible to review.

I see various improvements, bug fixes and so which I would agree on, but they are lost in an enormous amount of things which I do not understand what they do an why.

I fear this is strongly related to porting all mpes features here, which now heavily dominates the whole of SED. While I was the first to push for merging functionalities of MPES and hextof-processor, I did not have a full copy-paste in mind. 

I cannot refuse 4 months of work accumulated in here, and even if I believe these are all improvements to the current state of the code, I have no idea how most of this works, or what it does.

rettigl:
I agree this has grown quite large again. However, for me this was the most practical way to move forward with the functionalities I intended to add, as some parts also build on others. Alternatively, if I create a pull request for every commit, this would end up in a nested tree of branches and pull requests, where nobody has an overview anymore.
If you look at the individual commits, and the commit messages, it should become clear what is contained in this PR.

sed/loader/base/loader.py
rettigl: In principle I agree, however in this case I don't really know what to implement in the abstract class. You can make some suggestions in a new PR yourself.

rettigl: When looking into this, I realized that fixing the autodoc will require some more work, so I will even do only this in a separate PR.

rettigl: I created the PR #106 with updated docstrings and working documentation.

rettigl:
If reviewing the new code turns out difficult, I suggest you just have a look at the changes to the existing (core, binning) code. As everything I am currently doing branches off this branch, I really would like to merge it, and not grow it even further.

steinnymir (approved):
I understand you need this merged so you can continue growing it.
You can go ahead and merge this, as changes on the shared parts of code are fine by me.

We still need to better define how we want to include all this code from MPES in a way which works also outside of your labs though. But this needs to be done in a separate PR anyway.

zainsohail04 (approved):

PR #94, opened 2022-11-27
90 loader plugin api

rettigl:
This is my proposal for a plugin-based loader architecture, based on code from https://github.com/nomad-coe/nomad-parser-nexus


sed/core/processor.py
```diff
@@ -59,18 +59,25 @@ def __init__(
...
         self._attributes = MetaHandler(meta=metadata)
+
+        loader_name = self._config["core"]["loader"]
+        self.loader = get_loader(
+            loader_name=loader_name,
+            config=self._config,
+        )
+
         self.ec = EnergyCalibrator(
+            loader=self.loader,
```
zainsohail04, line 70: wouldn't it make sense to not load dataframe seperately for the calibrator? Giving it an input of a dataframe could work I believe. Right now, a dependency on the loader module has been created on calibrator module.

sed/loader/dask/loader.py
```diff
@@ -0,0 +1,104 @@
...
+from typing import Sequence
+from typing import Tuple
+
+import dask.dataframe as ddf
+
+from sed.loader.base.loader import BaseLoader
+from sed.loader.utils import gather_files
+
+
+class DaskLoader(BaseLoader):  # pylint: disable=too-few-public-methods
```
zainsohail04, line 19: I think renaming this module to something like GenericLoader makes more intuitive sense for the user for the features this loader supports.

sed/loader/utils.py
```diff
@@ -0,0 +1,45 @@
+"""Utilities for loaders
+"""
+from glob import glob
+from typing import cast
+from typing import List
+
+from natsort import natsorted
+
+
+def gather_files(
```
zainsohail04, line 10: My proposal was to have these utility functions in the base class [even above the abstract class], including the other dataframe features we wanted for our dataframe class. But perhaps I can later implement that, as it should not change the functionality, but hopefully just make the plugin-development process simpler.

zainsohail04 (changes requested):
Hi Laurenz. Thanks for providing your proposal and implementation. This is very similar to what I was proposing in the meeting so I have no major issues with it. What I might suggest different, I can later implement and make a PR for.
This structure should greatly ease the burden of implementing loaders for different instruments/sources.

sed/core/processor.py
rettigl: With the code as it is, the energy calibrator needs to access the files directly to extract the bias voltage. This can be changed in the future, but in this PR I wanted to concentrate on one thing.
The reason for handling the energy calibration data seperately in the energycalibrator module is that these are separate data from the ones loaded into the processor. One could off course also handle the loading/binning there...

sed/loader/dask/loader.py
rettigl: This can be done.

sed/loader/utils.py
rettigl: Well, this is so far only a function to collect files from a folder. How to integrate the dataframe class Steiin has been working on is probably a separate topic.

rettigl:
Thanks for your review. I think apart from these details, what we should discuss is the minimum requirement of a loader. I included now the use cases we had in mpes (loading from files, loading from folders), but nothing like loading from run-number or proposal, etc. Maybe this should be added to the interface if your require this.

zainsohail04:
> Thanks for your review. I think apart from these details, what we should discuss is the minimum requirement of a loader. I included now the use cases we had in mpes (loading from files, loading from folders), but nothing like loading from run-number or proposal, etc. Maybe this should be added to the interface if your require this.

I'd suggest you merge and I will use these updates to create a flash plugin [was already in the process after previous PR]. I think that'll make details clear on what is necessary.

sed/loader/utils.py
zainsohail04: Right. I think that'll hopefully reveal itself with more loader modules.

Issue #95, opened 2023-01-27
Binning with Tuple produces unexpected behavior


zainsohail04:
**Describe the bug**
The documentation states using a tuple with 3 parameters should be the start, stop and step_size. But currently, it is assuming it to be the num_steps.

**To Reproduce**
```
axes = ['dldPosX', 'dldPosY']
bins = [(430,900,1), (430,900,1)]
res = sp.compute(bins=bins, axes=axes)
```

**Expected behavior**
The behavior should be tuple(start, stop, step_size). Test currently tests tuple(start, stop, num_steps).

steinnymir:
User experience at HEXTOF pushed us to set the preferred triplet to be start, stop, step_size, so this is what I would stick with.
I see however the fast binning method is designed with number_of_steps, so if someone has strong arguments for it, I'm ok with this too.

An alternative option would be to reduce the options for input and ask each as explicit argument, e.g. `bin_dataframe(start=0,stop=100,step=1)` and `bin_dataframe(start=0,stop=100,number_of_steps=101)` should give equivalent outputs.

rettigl:
I suppose that we make this consistent for the actual binning functions, and implement different behavior in the class wrapper functions. I find giving the number of points to be more natural, as you have better overview over the amount of data you create...

zainsohail04:
Using the number of steps in backend makes good sense, then. We just have to have the wrappers designed in such a way that it's clear to the user what they are doing (like Steinn suggested).

I did a fix of the bug in branch 'flash' for now. Since users are already using it, I'll make a PR with that specific change.

rettigl:
We should discuss again what is the most natural thing to expect from the binning function to do.
I would opt for simulating a behavior that you would also have with a program that scans an axis:
if you tell it to start at A, end at B and do N steps, the natural expected behavior would be that the stepsize is (A-B)/N, and that the scan has indeed N+1points, to include the endpoints:

Example: 0, 10, 10 => [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]

This is at least how all scanning programs I know work like.

If we want to reproduce this with the binning, I would also expect these points to be the center coordinates of each bin. This results in N+2 bin edges, with the same stepsize and shifted by half a bin:

Example: 0, 10, 10 => [-.5, .5, 1.5, 2.5, 3.5, 4.5, 5.5, 6.5, 7.5, 8.5, 9.5, 10.5]

[...]

PR #98, opened 2023-01-31
Binning bug

zainsohail04:
Just setting up PR for #95 so we can figure this out. I came with a rough fix that shouldn't be used.

Note: Build will fail since test is testing num_steps.

### DIFF
--- a/tests/test_binning.py
+++ b/tests/test_binning.py
@@ -298,8 +298,9 @@ def test_simplify_binning_arguments(
         axes_,
         ranges_,
     )
-    #    for (bin_, bin__) in zip(bins__, bin_centers):
-    #        np.testing.assert_array_equal(bin_, bin__)
-    assert bins__ == bins_expected
-    assert axes__ == axes_expected
-    assert ranges__ == ranges_expected
+
+    for i, bin_ in enumerate(bins__):
+        np.testing.assert_array_equal(bin_, bins_expected[i])
+        np.testing.assert_array_equal(axes__[i], axes_expected[i])
+        if ranges__ is not None:
+            np.testing.assert_array_equal(ranges__[i], ranges_expected[i])
--- a/tests/test_binning.py
+++ b/tests/test_binning.py
@@ -1,8 +1,11 @@
 """This file contains code that performs several tests for the sed.binning module
 """
+from typing import Any
+from typing import Dict
 from typing import List
 from typing import Sequence
 from typing import Tuple
+from typing import Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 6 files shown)


rettigl:
I implemented now a version close to what I discussed in #95. I did not include the extension of the number of bins by one, but this can be easily added. What it does not is the following:
- For ranges, bins, it returns bin centers, aligned at (left range, ... right range except last bin).
Example: (0, 10), 10 gives [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] as bin _centers_.
- For bins as arrays, they are interpreted as bin centers, and converted into edges before binning.
I also cleaned up the simplify_binning_arguments function, and wrote a test function covering all cases. I removed the case that bins can be a single tuple of bins and ranges, because I don't see a purpose for it (if you have multiple axes, they will most certainly never have the same ranges...).

steinnymir (approved):
Again the same problem... 40 files changed.
This PR is about the bug in the binning function, so why does it refactor almost all docstrings of all functions? are these really changes here or do they come from merging other PRs in here?

The code changes I was able to find looked good. But I am certain I missed some.

For the sake of moving on, I will accept this. But lets try keep bug fix PRs to the point.

comments:
- I agree with the removal of the tuple method in the binning arguments. I added far too many options, but that is the default from numpy...
- for other definitions, we can continue our discussion in #95 and make smaller, targeted PRs eventually.

rettigl:
> Again the same problem... 40 files changed.

My apologies, this is a mistake in rebasing/merging of upstream I apparently made, which overwrites a lot of recent changes in master. Thanks for spotting this. Rebasing and force pushing should solve this, I believe.

rettigl:
Now it is properly rebased, I believe.

tests/test_binning.py
```diff
@@ -199,44 +201,105 @@ def test_bin_edges_to_bin_centers():
...
-def test_simplify_binning_arguments_tuple():
-    """Test the result of the _simplify_binning_arguments functions for bin tuples"""
-    bin_tuple = [tuple((r[0], r[1], b)) for r, b in zip(ranges, bins)]
-    bins_, axes_, ranges_ = _simplify_binning_arguments(bin_tuple, axes)
-    assert bins_ == bins
-    assert axes_ == axes
-    assert ranges_ == ranges
+    #    for (bin_, bin__) in zip(bins__, bin_centers):
+    #        np.testing.assert_array_equal(bin_, bin__)
+    assert bins__ == bins_expected
```
steinnymir, line 303: why not use `np.testing.assert_allclose` here? or if they need to be identical: `np.testing.assert_array_equal`

steinnymir (approved):
Nice! now I managed to read through easily! 😄 
looks good to me!

tests/test_binning.py
rettigl: They are not arrays, but lists of int or np.array at that point. Not sure if that will work, but I can try. Also not sure what the == actually tests, though.

steinnymir:
I think this is a good definition. It kind of sounds pythonic, when you think of how `range` works for example. We could have this as default behaviour, and add the endpoint argument to it allowing to include the last point when really necessary.
I believe there will be very few cases in which this would be used, but almost certainly someone who will complain if it is not there...

tests/test_binning.py
rettigl: Updated to assert_array_equals now

PR #104, opened 2023-02-21
WorkflowManager

steinnymir:
Here is a very basic implementation of the workflow manager as I described in the last meeting.

A notebook shows the basic principle of how this works and how a workflow step needs to be defined.

There is clearly more functionalities to be added, and I will continue adding some workflow steps in the coming days.

I'll make it a proper pull request once a few more workflow steps are defined.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@
 *.nx
 *.nxs
 
+# notebooks for development
+dev_*
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/sed/core/metadata.py
+++ b/sed/core/metadata.py
@@ -18,12 +18,14 @@ class MetaHandler:
         # TODO: #35 add pretty print, possibly to HTML
         return str(self._m)
 
-    def add(self, v: Dict, duplicate: str = "raise") -> None:
+    def add(self, v: Dict, category=None, duplicate: str = "raise") -> None:
         """Add an entry to the metadata container
 
         Args:
             v: dictionary containing the metadata to add.
                 Must contain a 'name' key.
+            category: sub-dictionary container where to store the entry. e.g. "workflow"
+                for workflow steps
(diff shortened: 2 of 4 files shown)


steinnymir:
Thank you @rettigl for looking into this. I'll try answer your questions in a few points
1. this is intended as an automatic pipe, which runs all provided methods with the given parameters with no need for user input. This is necessary if we want this to run as an automatic correction system (even though only on the long run). Also, makes it the simplest approach for users, as they need only call a list of trnsformations, which can be pre-defined, making for example a standard correction pipe for a given beamtime/experiment.
2.  The definition of the parameters should be a separate problem. Each user should have the freedom of using the tool that they prefer to do this. However, we of course will provide one (python based) method to determine them for each correction method we provide. The structure of these is not yet defined, but could be presented in jupyter notebooks as examples. This also gives a space to properly describe how these transformations work, should a user be interested. The generated parameters can be then stored in a json/config file
3. this json/config currently these methods need explicit parameters, but this can easily be switched to picking values from a json/config file instead. These could be defined as in point 2, or kept as fixed values provided by the beamline (for corrections which don't change over an experiment)
4. I know the function is very similar to what you added, but if you look at the structure of it, (class) it does not have the same implementation. Obviously, we should not keep processor and workflow at the end, and the corrections you imported from MPES will need some adaptation for the new system.
5. I am also planning to make a workflow_step generator function, which passing it a function, returns a subclass of `WorfklowStep` which runs the given function. THis could both simplify transition of existing mehtods, and add the possibility of user generated transformations

I hope this wall of text answers some of your concerns, and keep in mind this is still work in progress 😄

zainsohail04:
as discussed, I will try to include #97 in this (just for my reference)

steinnymir:
This PR could probably be closed, as I have restarted from an other more recent branch, to avoid needing to merge in all changes made in the mean time.

Other than that there is no real update so far, I hope to have something soon though.

PR #105, opened 2023-02-24
Simpler sed version

daviddoji:
As promised, this is an attempt to simplify some of the functionality of the SED library.

There is also a notebook showing the functionality.

### Known limitations

- it's supposed to be run in Maxwell, as it loads in memory the whole directory of h5 files.
- there is also a hardcoded dictionary from where to read the data from the files. 

All of that can be easily changed.


daviddoji:
> Any updates on including this in sed, @daviddoji? Perhaps also the SXP loader?

We're at full commission at SXP and I don't have time to write such loader. Maybe during the shutdown.

PR #106, opened 2023-03-05
Autodoc

rettigl:
This PR contains fixed configurations for autodoc documentation, and updates all docstrings to a common style compatible with sphinx. It branches off the workflow_improvements, so this should be merged first.


rettigl:
We might want to add a workflow for automatic generation and publishing the docs e.g. on an guthub.io page.

sed/binning/binning.py
```diff
@@ -43,34 +44,35 @@ def bin_partition(
...
-            The order will be the order of the dimensions in the resulting array.
-        ranges: list of tuples containing the start and end point of the binning range.
-        histMode: Histogram calculation method. Choose between
-            "numpy" which uses numpy.histogramdd, and "numba" which uses a
-            numba powered similar method.
-        jitter: a list of the axes on which to apply jittering.
-            To specify the jitter amplitude or method (normal or uniform noise)
-            a dictionary can be passed.
-            This should look like jitter={'axis':{'amplitude':0.5,'mode':'uniform'}}.
+        part (Union[dask.dataframe.core.DataFrame, pd.DataFrame]): dataframe on which
```
steinnymir, line 47: Does autodoc need to have all types explicitly mentioned in the comment?
It seems dangerous to leave to the programmer to change the data type comment if the type hint is changed.

tutorial/3 - Conversion Pipeline for example time-resolved ARPES data.ipynb
```diff
@@ -19,12 +19,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "%load_ext autoreload\n",
```
steinnymir, line 22: The notebook should be clean when pushed, outputs should not be included.

steinnymir (changes requested):
Nice to see the documentation building up! thanks for looking into this!
I left a couple of comments to address first though

sed/binning/binning.py
rettigl: This is the difference of the two versions:
With typing in docstring:
image
Without typing in docstring:
image

We can change that for sure again, if you think the latter is better.

steinnymir: You have a good point. Would be so nice to use the python 3.10 syntax for type hints instead, but it unfortunately breaks retro-compatibility. 
Do you think there might be a git workflow step which checks that type hints match the comments?

steinnymir: What happens if there is a mix of both cases? As, for example, if we do not update the docstring of a function. Will sphinx add it where it is missing? or will it add it everywhere if we toggle it on?

rettigl: If they are missing, they are filled in as shown in the example. I don't know what happens if they are contradicting, I suppose the ones in the docstring win...

steinnymir: That would be perfect then. We can modify the ones we notice, and the others will have a less readable, yet complete, documentation!

steinnymir (approved):

PR #108, opened 2023-03-14
Mpes metadata

rettigl:
This PR implements metadata collection and conversion to NeXus for mpes.


rettigl:
This should be ready for review. It's again a rather large PR, because I restructured a few things for collecting the required metadata for Nexus conversion in a consistent way.

steinnymir (commented):
We discussed this already in the last PR, this is too big to review. I see potentially good things, but I have no intention of going through thousands of lines of changes to understand them. Please do not mix additions, fixes and small modifications in a single PR...

rettigl:
How whould we move forward with this? I currently build everything I do based on this, which makes the backlog even larger. Thus, i would like to merge this soon. Howe we progress with the other PRs, like the complete restructuring, should also take these developments into account, this I suggest to merge first (also the developments on the flash reader should take the changes to the reader interface implemted here into account, if you agree to the structure).

zainsohail04:
I believe since we started this project with poetry, it'd be great if we keep all our development with that. That should be the first place where dependencies should change and requirements.txt should be secondary.
A lot of users found the poetry experience very pleasant in comparison to using conda so I would really urge that we all try to also use it.

Regarding the PR, I think the example notebook should only include a basic overview that highlights how the metadata is handled. The processing parts are already covered in *3* so they could be removed.

I've taken a look at this PR before and now and checking the new functionality, everything seems to work. Though Steinn's comment is valid, a lot of files need to change to make the metadata handling with sed work.

zainsohail04 (approved):

rettigl:
> I believe since we started this project with poetry, it'd be great if we keep all our development with that. That should be the first place where dependencies should change and requirements.txt should be secondary.
> A lot of users found the poetry experience very pleasant in comparison to using conda so I would really urge that we all try to also use it.

Do you mean that there are some inconsistencies between requirements.txt and poetry.toml here? That might be true, I will fix this.

> Regarding the PR, I think the example notebook should only include a basic overview that highlights how the metadata is handled. The processing parts are already covered in 3 so they could be removed.

Also a valid comment. I will reduce the metadata notebook to the bare conversion with parameters from the config.

rettigl:
I updated pyproject.toml, and removed requirements.txt, because this is redundant, I was not aware of this. I also stripped down the example notebook.
However, I could not get the project to install with poetry, it just runs forever in loops (I stopped after ~20 minutes). Maybe there is some inconsistency between requirements of packages? Put then poetry should be able to tell, no? I don't understand this.

steinnymir (approved):
I see how this PR has a lot of major changes which affect your future development @rettigl. I think this should be merged so we can move on.

rettigl:
@daviddoji @zainsohail04 I am not getting poetry to install and generate a new poetry.lock file. Even if I do it on the master branch, it does not work if I remove the poetry.lock file, it just takes forever to resolve dependencies. Shall I just merge and remove the poetry.lock file, or do you have any advice, or could generate and push the file for me?

rettigl:
> @daviddoji @zainsohail04 I am not getting poetry to install and generate a new poetry.lock file. Even if I do it on the master branch, it does not work if I remove the poetry.lock file, it just takes forever to resolve dependencies. Shall I just merge and remove the poetry.lock file, or do you have any advice, or could generate and push the file for me?

Found it, the version of sphinx was apparently kept back at a too old version... (?). At least lifting this restriction helped.

daviddoji:
> @daviddoji @zainsohail04 I am not getting poetry to install and generate a new poetry.lock file. Even if I do it on the master branch, it does not work if I remove the poetry.lock file, it just takes forever to resolve dependencies. Shall I just merge and remove the poetry.lock file, or do you have any advice, or could generate and push the file for me?

`poetry.lock` file is just a reference that contains hashes of the libraries that were installed using the dependencies of the `pyproject.toml`, which is the reference for correct build instructions.

If dependencies are pinned in `pyproject.toml`, one can forget about the `poetry.lock` file. 

Given the amount of dependencies of `sed`, maybe is not a bad idea to freeze them all and commit the lock file to VC. Doing it so, one will avoid this slowness in resolving dependencies (only the first time has to be done) and the same version of libraries will be installed for everybody.

steinnymir:
> As such, we should keep the version restictions as small as possible, to allow it to be compatible with as many other packages as possible.

I fully agree

daviddoji:
> It's clear that the poetry.lock just describes the current state, but it should be consistent with the pyproject.toml, and more importantly poetry should be able to find a solution. Before lifting the fix on sphinx, it did not so for me for over two hours...
> 
> We discussed this before: We want to build a package here that people can use as part of their data analysis environment, not a stand-alone project. As such, we should keep the version restictions as small as possible, to allow it to be compatible with as many other packages as possible.

I won't discuss again the benefits of using a tool like poetry here. Lock files are written after the `pyproject.toml` has been resolved, dependency-wise. If this takes too long, it means that there is an intricated set of dependencies that it is difficult for poetry to solve. If it does, a lock file is written, otherwise the lock file is not generated.
As it was generated in the past (it was even commited to the repo), I don't get why it should not be generated once again using the same version especified in the lock file.

On the other hand, I agree with you that pinned dependencies should be kept to the minimum. What I don't understand is why sphinx is a mandatory dependency for sed, and not an extra one.

PR #110, opened 2023-03-22
notebook now working

zainsohail04:


sed/core/parameter_generation.py
```diff
@@ -36,8 +35,7 @@ def get_preprocessing_step(self):
         return self.PREPROCESSING_STEP(**self.parameters)
 
     def metadata(self) -> dict:
-        metadict = {"version": __version__}
-        return metadict
+        pass
```
steinnymir, line 38: I had just changed this, but its placeholder anyway so its fine to revert it

steinnymir (approved):

PR #111, opened 2023-03-23
config file simplified and unnecessary module removed

zainsohail04:


rettigl:
I'm fine with renaming / removing this as a module. However, please also fix the test, take care of linting, and update the docs.

sed/__init__.py
```diff
@@ -1,7 +1,7 @@
 """sed module easy access APIs
 
 """
-from .core.processor import SedProcessor
+# from .core.processor import SedProcessor
 
 __version__ = "0.1.0"
-__all__ = ["SedProcessor"]
+# __all__ = ["SedProcessor"]
```
rettigl, line 7: Why do you want to remove this? This is typically the only thing you want to import and use (right now).

sed/core/config.py
```diff
@@ -1,60 +1,39 @@
...
 import json
-import os
-from importlib.util import find_spec
 from pathlib import Path
 from typing import Union
 
 import yaml
 
-package_dir = os.path.dirname(find_spec("sed").origin)
+DEFAULT_CONFIG_PATH = Path(__file__).resolve().parent / "default.yaml"
```
rettigl, line 9: I find it more systematic to base everything onto the base folder of the package (package_dir), rather than the individual files of the package...)

sed/core/config.py
```diff
@@ -1,60 +1,39 @@
...
-    config: Union[dict, str] = None,
-    default_config: Union[
-        dict,
-        str,
-    ] = f"{package_dir}/config/default.yaml",
-) -> dict:
-    """Load the config dictionary from a file, or pass the provided config dictionary.
+def parse_config(config: Union[dict, str] = None) -> dict:
+    """Load the default config dictionary, and update it if a user file or dict
+    is provided.
```
rettigl, line 14: This version is not reusable anymore. What I discussed with Steinn at DPG is that we want most likely for every scan to have a config file that is read automatically, and merged with a per-site config, that in-turn is merged with a per-package config. Thus the requirement of reuasability of this function....

sed/core/default.yaml
rettigl, line 1: Best put it into a config directory in sed/. We can also put other example configs there.

tests/test_config.py
```diff
@@ -70,12 +71,16 @@ def test_load_config_raise():
...
     """Test the merging of a config and a default config dict"""
-    dict1 = {"key1": 1, "key2": 2, "nesteddict": {"key4": 4}}
-    dict2 = {"key1": 2, "key3": 3, "nesteddict": {"key5": 5}}
-    dict3 = insert_default_config(config=dict1, default_config=dict2)
-    assert isinstance(dict3, dict)
-    for key in ["key1", "key2", "key3", "nesteddict"]:
-        assert key in dict3
-    for key in ["key4", "key5"]:
-        assert key in dict3["nesteddict"]
-    assert dict3["key1"] == 1
```
rettigl, line 81: See above. The reusability should remain.

rettigl (changes requested):
I am fine with renaiming the module, but not with changing its functionality. What is the reason for it? I very much opt for keeping the resusability of the functions, as we might want to combine different levels of config in the future.

sed/__init__.py
steinnymir: I also have this always commented out but the reason is that it forces importing everything, due to the current structure of the processor. On the long run, this should be there though

sed/core/default.yaml
rettigl: You can include any files/folder in a MANIFEST file or so, no?

rettigl: This is an excellent idea. Do you know how to set this up?

steinnymir: no clue, and certainly will make things more complicated with different OS. I'll set up an issue for this to start with!

tests/test_config.py
rettigl: Fixed by my revision

sed/__init__.py
rettigl: I re-enabled this here. If we want to disable this, let's have a discussion on this, and a separate PR.

sed/core/config.py
rettigl: I reverted to the old systematics, and kept the config folder containing the default config. It might also contain example config files for the different experiments in the future.

sed/core/config.py
rettigl: I reverted to a reusable version, and renamed it. It is indeed reused heavily in the new code.

sed/core/default.yaml
rettigl: Implemented in the current version. For packaging, we still need to work on documentation, and how to install files into the user space.

rettigl (approved):

rettigl:
I updated this PR to include loading functionality for a hierarchy of config files, from the given one, to a file in the current folder, to a user-file in the user's home directory, down to the default config. 
I also included functionality to write out the calibration and correction parameters to the per-folder config file, as discussed in issue #121 
@steinnymir @zainsohail04 please re-review

sed/core/config.py
```diff
@@ -0,0 +1,222 @@
...
+
+package_dir = os.path.dirname(find_spec("sed").origin)
+
+
+def parse_config(
+    config: Union[dict, str] = None,
+    folder_config: Union[dict, str] = None,
+    user_config: Union[dict, str] = None,
+    system_config: Union[dict, str] = None,
+    default_config: Union[
```
zainsohail04, line 20: I think the difference between the three configs isn't quite clear, at least to me. So I can imagine it'd be likewise for the users.

sed/core/config.py
```diff
@@ -0,0 +1,222 @@
...
+        str,
+    ] = f"{package_dir}/config/default.yaml",
+    verbose: bool = True,
+) -> dict:
+    """Load the config dictionary from a file, or pass the provided config dictionary.
+
+    Args:
+        config (Union[dict, str], optional): config dictionary or file path.
+                Files can be *json* or *yaml*. Defaults to None.
+        folder_config (Union[ dict, str, ], optional): folder-based config dictionary
```
zainsohail04, line 31: Like what do folder-based values means? and why is it folder config

sed/core/config.py
```diff
@@ -0,0 +1,222 @@
...
+            or file path. The loaded dictionary is completed with the folder-based values,
+            taking preference over user, system and default values. Defaults to the file
+            "sed_config.yaml" in the current working directory.
+        user_config (Union[ dict, str, ], optional): user-based config dictionary
```
zainsohail04, line 35: This one makes sense

sed/core/config.py
```diff
@@ -0,0 +1,222 @@
...
+            or file path. The loaded dictionary is completed with the user-based values,
+            taking preference over system and default values.
+            Defaults to the file ".sed/config.yaml" in the current user's home directory.
+        system_config (Union[ dict, str, ], optional): system-wide config dictionary
```
zainsohail04, line 39: And how are these default values installed to userspace?

sed/core/config.py
```diff
@@ -0,0 +1,222 @@
...
+        if Path(user_config).exists():
+            user_dict = load_config(user_config)
+            if verbose:
+                print(f"User config loaded from: [{str(Path(user_config).resolve())}]")
+
+    system_dict: dict = None
+    if isinstance(system_config, dict):
+        system_dict = system_config
+    else:
+        if system_config is None:
```
zainsohail04, line 92: generally I like this idea of system config

sed/core/config.py
rettigl: Ill update the docu and provide examples

rettigl:
@zainsohail04 please check if the updated documentation makes now sense to you...

zainsohail04:
> @zainsohail04 please check if the updated documentation makes now sense to you...

Thank you for the updates. It's all understandable now.

PR #114, opened 2023-03-31
Flash loader

zainsohail04:
I am putting this branch for PR so we can soon merge it

The loader is quite complex by nature of the machine, hence the testing will also include such. 

It is mostly a porting from hextof-processor with some bugfixes, a change in the multiprocessing backend, a bit of refactoring. 

I will keep pushing the tests as I do them but I'd suggest you to review the loader first.


rettigl:
I am happy to review that, however, before please
-fix the merge conflicts first (either rebase onto main (preferrred), or merge main into this branch)
-make sure linting and tests pass
-provide some test data and an example notebook
 WIthout these, it's quite difficult to review, and won't be mergabale anyways

steinnymir (commented):
Are these requirements actually relevant for the flash loader? most of them look like the mpes requirements, and if not relevant to the additions in this PR, I would leave them out.

daviddoji:
I agree with @rettigl here, at least regarding the first and last point. Sooner or later those requirements have to be met in order to merge this branch. So, instead of review some not-ready-to-merge code, I would address them as soon as possible.
This will save time in the long run to the reviewers ;)

Issue #101, opened 2023-02-07
Possibility for normalization by the number of FEL pulses:


yacremann:
We need a possibility to normalize the binned signal by the number of FEL pulses or the intensity monitor of the free electron laser. 

**Preferred solution:**
Please add a way to specify that the SED processor can be operated in the per-electron mode (standard), or a new "per-pulse"-mode. This way, we could instantiate a second processor in per-pulse-mode, which can be used for normalization. 

We also considered to include an additional 'per pulse'-dataframe in SED, but this would significantly alter the structure of the code and will reduce the performance if normalization is not needed.

rettigl:
I'm not 100% sure if that works, but it appears to me that you can select the index via the config parameter dict "channels"

self.all_channels: dict = self._config.get("channels", {})

It should be straight forward to create two instances with different configs

steinnymir:
@yacremann I believe you would want the old ddMicrobunches table, right?
If I am not mistaken, this can be obtained from the current dataframe by droppping the electron index, and leaving only the per-pulse channels, as @rettigl correctly suggests
I can look into making a simpler access to generating this, as is it indeed useful to normalize FEL intensity, but also delay stage positions (less crucial with the current laser system).

rettigl:
I was also thinking about this in the context of lab experiments. In principle, such a normalization histogram could be derived from a per-electron dataframe, if every electron has a timestamp, which could be generated for the FLASH data, and is already implemented for our data.

zainsohail04:
We can provide a notebook with said idea of filtering, or it can be part of the workflow within the context of workflow manager (makes more sense).

steinnymir:
a notebook or example of the exact use case for the required feature is always greatly appreciated!

If I remember correctly, the most use of the per-pulse dataframe consisted in binning 1D traces in parallel to the final binning array to use as normalization arrays. This should be easy to implement in the workflow-manager/processor class.

rettigl:
I have thought about this a bit. In principle, maintaining two dataframes seems a bit redundant and a good design choice. On the other hand, deriving e.g. a "time per electron" column out of time stamps or bunch numbers probably creates a substantial amount of overhead, so maybe the former approach might perform much better...

yacremann:
@rettigl I am also not sure that generating a per-pulse dataframe from the per-electron dataframe is efficient. In addition, it will be necessary to remove duplicates, which is likely slow.
[...]

PR #116, opened 2023-04-03
Histograms from timed dataframe

rettigl:
This adds support for a per-time-unit dataframe, and histogram calculation based on this dataframe.

Histogram calculation in the example notebook takes on our machine ~55s. This can potentially still be improved, e.g. by taking our fast binning method (currently fails because of xarray generation and concatenation issues)


rettigl:
This does not produce the right kind of normalization histogram yet, I think...

rettigl:
This implements both variants now. Both routes produce almst identical results, but histograms from timed dataframes are substantially faster. Also a normalization routine for the compute function is added.

One issue remains: If histogram axes are jittered, the actual values of the jitter are different during computation of the data and of the normalization histogram, leading to small errors in the histogram on the order of a percent or so. A workaround would be to include the histogram calculation into bin_partition. This would also avoid reading the data twice, which I think is the major time bottleneck at the moment.

rettigl:
Histogram ontop of data (continously scanned delay stage):
grafik
Performance comparison:
grafik
Two histogram methods compared (bottom: difference):
grafik
Histogram computed twice and compared (jittered):
grafik

Bottomline: All effects of randomness and difference between methods are < 1%.

rettigl:
@zainsohail04 This one is pending on your implementation into the Flash loader (and some review)...

rettigl:
Rebased, and added dummy implementation for flash loader. Tests are also still pending.

rettigl:
@zain-sohail For me the flash timed_dataframes don't work. They consume endless amounts of memory and never finish computation.

zain-sohail:
> @zain-sohail For me the flash timed_dataframes don't work. They consume endless amounts of memory and never finish computation.

Screenshot 2023-10-10 at 17 25 35
Just tried and it seems to work fine. Where are you facing a problem?

rettigl:
After merging the fast loading code from main, it is at lead performing faster now, but still does not seem to do the right thing:
I am loadings runs=[44824, 44825, 44826, 44827]
grafik
grafik
It does not make sense that the timed dataframe has 8 times _more_ rows...

rettigl:
Or does it? Are there so many empty macrobunches?

rettigl:
Hmm, okay the generated histogram does indeed not look to bad
grafik

rettigl:
The time-stamp version does produce some strange artefacts, though:
grafik

steinnymir:
> Or does it? Are there so many empty macrobunches?

if you look at the tables you show, the pulse IDs in the electron dataframe are very sparse (7, 38, 69...) so it seems like count rate was really bad, and therefore I expect many empty pulses.

rettigl:
rebased to current main

rettigl:
> > Or does it? Are there so many empty macrobunches?
> 
> if you look at the tables you show, the pulse IDs in the electron dataframe are very sparse (7, 38, 69...) so it seems like count rate was really bad, and therefore I expect many empty pulses.

I realized that after posting my previous post, but don't undertand the data format completely yet. What exactly are the Train, pulse, and electron IDs? Are these Macrobunches, microbunches, and electrons per microbunch? And what is the time-base for the timed_dataframe? I was expecting macrobunches, but then I would expect one row per trainID, no? Now it seems one row per microbunch. This is certainly typically more than electrons. And why is it going up to 1000, if tpyically there are 400 or 500 microbunches in a macrobunch?

steinnymir:
How you described it is all correct.
the reason there are 1000 is I think because the data structure accomodates up to 1000 pulses, as flash can theoretically provide 800, which it never does, max 500 in reality.
I'm not sure if using trainID (macrobunches) would be still ok, don't you think a coarser step size induces more noise?

rettigl:
> I'm not sure if using trainID (macrobunches) would be still ok, don't you think a coarser step size induces more noise?

All you want here is to capture changes of the scanned variable, which typically is anyways only read out once per macrobunch, no? So no, there should be no difference. It would only make a (marginal) difference for parameters that are varied within the macrobunch. Not sure, though, how this would look like if you consider, e.g. BAM correction etc.

steinnymir:
> All you want here is to capture changes of the scanned variable, which typically is anyways only read out once per macrobunch, no? So no, there should be no difference. It would only make a (marginal) difference for parameters that are varied within the macrobunch. Not sure, though, how this would look like if you consider, e.g. BAM correction etc.

Right! but still there might be some pulse resolved normalizations one would want, like the FEL intensity for example. I'm afraid we need to keep the per-pulse dataframe for those

zain-sohail:
> After merging the fast loading code from main, it is at lead performing faster now, but still does not seem to do the right thing: I am loadings runs=[44824, 44825, 44826, 44827] 

Wouldn't make sense that there is a timing difference in the compute of dataframe, only makes a difference in creating buffer files quicker, I would imagine. Regarding your question about why the dataframe is bigger, I think Steinn managed to answer it.

sed/core/processor.py
```diff
@@ -1198,6 +1298,12 @@ def add_jitter(self, cols: Sequence[str] = None):
             cols=cols,
             cols_jittered=cols,
         )
+        if self._timed_dataframe is not None:
```
steinnymir, line 1301: This does not work when jittering on columns which are not in the timed dataframe.
With HEXTOF, jittering is done exclusively on dld parameters, which are all by electron and therefore do not enter thie timed_dataframe. I would suggest either removing this or setting a check for only applying jitter on columns present in the timed_dataframe.

sed/core/processor.py
```diff
@@ -1112,6 +1194,12 @@ def append_energy_axis(
                 calibration=calibration,
                 **kwds,
             )
+            if self._timed_dataframe is not None:
```
steinnymir, line 1197: Same as for the jittering. Energy axis is electron based and should not be present in the timed_dataframe. I would remove this for energy calibration

sed/core/processor.py
```diff
@@ -582,6 +649,10 @@ def apply_momentum_correction(
             self._dataframe, metadata = self.mc.apply_corrections(
                 df=self._dataframe,
             )
+            if self._timed_dataframe is not None:
```
steinnymir, line 652: same as for energy and jitter

steinnymir (changes requested):
Looks good! I tested it together with the hextof changes from #169 and seems to work correctly. Only exception the comments I made. Once those are resolved you can merge for what concerns me.

steinnymir:
I accidentally uploaded the branch where i tested this. I'll leave it till this is merged. its `hist_testing`

sed/core/processor.py
rettigl: I implemented checks and exceptions for all these cases. Maybe one could extend the processor tests to be run for both mpes and flash test data, to include these exceptions in the tests. Could also be an issue, I think it works well.

rettigl:
I fixed this, and modified the tests to test the timed dataframes.

rettigl:
> I accidentally uploaded the branch where i tested this. I'll leave it till this is merged. its `hist_testing`

I also updated your test branch

steinnymir (approved):
I tested this and seems to work fine. LGTM!

steinnymir:
closes #101

rettigl:
I am working on improving tests for this still, and update the new processor functions, then I will merge

sed/calibrator/energy.py
```diff
@@ -1442,8 +1442,10 @@ def align_dld_sectors(
             dask.dataframe.DataFrame: Dataframe with the new columns.
             dict: Metadata dictionary.
         """
-        sector_delays = sector_delays or self.sector_delays
-        sector_id_column = sector_id_column or self.sector_id_column
+        if sector_delays is None:
+            sector_delays = self.sector_delays
+        if sector_id_column is None:
+            sector_id_column = self.sector_id_column
```
rettigl, line 1448: @steinnymir This construction produced an error for me, once you do not actually pass the same array (reference). Then the truth value of this comparison is ambigous. So I suggest to generally stick with comparison to None.

PR #117, opened 2023-04-03
Histograms from timestamps

rettigl:
This adds histogram calculation based on the timeStamp column in the dataframe

Histogram calculation in the example notebook takes on our machine close to 2 minutes. Additionally, binning becomes slower when timeStamps are generated, even if they are not used during binning.

### DIFF
--- a/sed/core/processor.py
+++ b/sed/core/processor.py
@@ -18,6 +18,7 @@ import psutil
 import xarray as xr
 
 from sed.binning import bin_dataframe
+from sed.binning.utils import bin_centers_to_bin_edges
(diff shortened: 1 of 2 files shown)


rettigl:
PR #116 and PR #117 are alternative implementations of normalization histograms, open for debate.

rettigl:
The current implementation still suffers from the bin-centers vs. bin-edges problem, I think. Needs further inspection.

rettigl:
Closing this, as it is implemented in PR #116 as well, and thus obsolete

PR #119, opened 2023-05-06
Point selector

rettigl:
This PR separates the feature selection process from the spline generation step, and adds a GUI interface for that. It is based off the metadata branch, so this should be merged first, and then this branch be rebased (then there are only 3 files changed).


rettigl:
I rebased onto main now, so it should be fine to review. Mostly adds the GUI. Splitting of this step prepares for separating this as a parameter generation step for the new workflow manager layout.

Issue #66, opened 2022-04-09
Numba binning not producing the same result when providing bin edges as list


rettigl:
**Describe the bug**
When the new binning function get the bin edges as a list, it does not produce the same result as if it gets the bins as number of bins and bin ranges. The version with bin edges as list seems to ignore the last edge (?).

**To Reproduce**
The following code executed in in the example data notebook reproduces the problem:

```
binAxes = ["t"]
nBins = [300]
binRanges = [(65000, 67000)]
res1 = bin_dataframe(
    df=ddf,
    axes=binAxes,
    bins=nBins,
    ranges=binRanges,
    histMode="numba",
    nCores=20,
)
binAxes = ["t"]
res2 = bin_dataframe(
    df=ddf,
    axes=binAxes,
    bins=[(65000, 67000,300)],
    histMode="numba",
    nCores=20,
)
binAxes = ["t"]
nBins = np.linspace(65000, 67000, 300, endpoint=True)
res3 = bin_dataframe(
    df=ddf,
    axes=binAxes,
    bins=nBins,
    histMode="numba",
    nCores=20,
)
res1.plot()
res2.plot()
res3.plot()
print((res1.t==res2.t).all())
print((res1==res2).all())
print((res1.t==res2.t).all())
print((res1==res3).all())
```
Output:
```
<xarray.DataArray 't' ()>
array(True)
<xarray.DataArray ()>
array(True)
<xarray.DataArray 't' ()>
array(True)
<xarray.DataArray ()>
array(False)
```
**Expected behavior**
The results should be identical, as the t-axes are identical

**Screenshots**
Graph:
![output](https://user-images.githubusercontent.com/53396064/162589247-a4e78d61-1127-415c-a165-869d84761a00.png)

zainsohail04:
Is this bug still there?

rettigl:
This bug still exists. Apparently, our tests for binning are very incomplete. In particular, there are not tests for bin_partition or bin_dataframe. 
Indeed, binning with provided bins as arrays fails right now, because the result array size is wrongly calculated. This at least is fixed in PR #98 
Additionally, I encountered that performance in binning with arrays seems to be substantially worse than with bins as int for high-dimensional data, and under certain circumstances binning hangs completely in an infinite loop. I expect this to happen in binsearch, but this needs further investigation.

rettigl:
[...]

PR #120, opened 2023-05-08
A PR to address the bug in numba binning

rettigl:
See #66 for details.
This initial commit contains additional test data to reproduce and test for the bug.

### DIFF
--- a/sed/binning/numba_bin.py
+++ b/sed/binning/numba_bin.py
@@ -55,7 +55,8 @@ def _hist_from_bin_range(
         is_inside = True
         flatidx = 0
         for i in range(ndims):
-            j = (sample[t, i] - ranges[i, 0]) * delta[i]
+            # strip off numerical rounding errors
+            j = round((sample[t, i] - ranges[i, 0]) * delta[i], 11)
             # add counts on last edge
             if j == bins[i]:
                 j = bins[i] - 1
--- a/sed/binning/numba_bin.py
+++ b/sed/binning/numba_bin.py
@@ -56,6 +56,9 @@ def _hist_from_bin_range(
         flatidx = 0
         for i in range(ndims):
             j = (sample[t, i] - ranges[i, 0]) * delta[i]
+            # add counts on last edge
+            if j == bins[i]:
+                j = bins[i] - 1
             is_inside = is_inside and (0 <= j < bins[i])
             flatidx += int(j) * strides[i]
             # don't check all axes if you already know you're out of the range
--- a/sed/binning/numba_bin.py
+++ b/sed/binning/numba_bin.py
@@ -81,6 +81,8 @@ def binsearch(bins: np.ndarray, val: float) -> int:
     Returns:
         int: index of the bin array, returns -1 when value is outside the bins range
     """
+    if np.isnan(val):
+        return -1
     low, high = 0, len(bins) - 1
     mid = high // 2
     if val == bins[high]:
--- a/tests/test_binning.py
+++ b/tests/test_binning.py
@@ -17,18 +17,26 @@ from sed.binning.utils import bin_centers_to_bin_edges
 from sed.binning.utils import bin_edges_to_bin_centers
 from .helpers import get_linear_bin_edges
 
-sample = np.random.randn(int(1e2), 3)
+sample = np.random.randn(int(1e5), 3)
 columns = ["x", "y", "z"]
 sample_df = pd.DataFrame(sample, columns=columns)
-bins: Sequence[int] = tuple(np.random.randint(5, 50, size=3))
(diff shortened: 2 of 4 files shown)


rettigl:
This PR addresses two bugs in numba binning from bins ranges:
- samples on the last bin edge are discarded, while they are included in the last bin in np.histogram and in numba from bins. Specific tests for this have been implemented, and these cases have been included.
- under rare circumstances, bin edges are slightly off due to numerical rounding errors. A test for such a case has been included, and a bufgix has been implemented by rounding bin numbers to 11 digits of precision. While this fixes the bug I have identified, and the other random tests do pass, I cannot guarantee there are no further cases where this might still fail. However, I find theres edge cases also not so relevant, as no events are getting lost, but just shifted between bins, which is anyways irrelevant when jittering is employed.

Performace is lightly affected, but only by a few percent, see benchmarks in #66 and the corresponding benchmark notebook, which is also added here.

sed/binning/numba_bin.py
```diff
@@ -55,7 +55,11 @@ def _hist_from_bin_range(
         is_inside = True
         flatidx = 0
         for i in range(ndims):
-            j = (sample[t, i] - ranges[i, 0]) * delta[i]
+            # strip off numerical rounding errors
+            j = round((sample[t, i] - ranges[i, 0]) * delta[i], 11)
+            # add counts on last edge
+            if j == bins[i]:
```
daviddoji, line 61: instead of comparing, what about something like this:
```
j = j - 0.5 * (j == bins[i])
```
the operation is then vectorized and we don't have to compare in every step

tests/test_binning.py
```diff
@@ -17,17 +17,35 @@
...
+
+sample_int = np.random.randint(
+    low=60000,
+    high=70001,
+    size=(int(1e5), 3),
+    dtype=int,
+)
+bins_int = tuple(np.random.randint(10, 300, size=3, dtype=int))
+ranges_int = [
+    tuple(np.sort(np.random.randint(60000, 70000, size=2, dtype=int))),
```
daviddoji, line 38: why repeating the same tuple 3 times?

sed/binning/numba_bin.py
```diff
@@ -55,7 +55,11 @@ def _hist_from_bin_range(
         is_inside = True
         flatidx = 0
         for i in range(ndims):
-            j = (sample[t, i] - ranges[i, 0]) * delta[i]
+            # strip off numerical rounding errors
+            j = round((sample[t, i] - ranges[i, 0]) * delta[i], 11)
```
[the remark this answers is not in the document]
rettigl: I'm not sure I understand what you mean. We do create the array in advance (line 45). The round here happens to strip off numerical round errors that happen for certain combinations of edge values. 
The underlying reason is that here we do the forward transformation (position -> bin index), while in np.histogramdd linspace is used to create the bin edges, i.e. (bin index -> position). These two directions can produce slightly different results due to numerical float rounding errors, which I strip off at the 11th digit with this round command, which emperically does the trick, at least for the test case I identified.

sed/binning/numba_bin.py
rettigl: We loop through each event, and each dimension individually (for t, for dim), so j is a single float number. I don't think that this will make a difference.

sed/binning/numba_bin.py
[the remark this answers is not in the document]
daviddoji: > I'm not sure I understand what you mean. We do create the array in advance (line 45). The round here happens to strip off numerical round errors that happen for certain combinations of edge values. The underlying reason is that here we do the forward transformation (position -> bin index), while in np.histogramdd linspace is used to create the bin edges, i.e. (bin index -> position). These two directions can produce slightly different results due to numerical float rounding errors, which I strip off at the 11th digit with this round command, which emperically does the trick, at least for the test case I identified.

Sorry, my bad. I read that `j` was an array instead of a single float. As I said, as `round` is now supported in numba, the change makes sense.

sed/binning/numba_bin.py
daviddoji: Same as above.

tests/test_binning.py
daviddoji: I understand the code. My point was about repeating three times the same line. Why not something like:
```
ranges_int = [tuple(np.sort(np.random.randint(60000, 70000, size=2, dtype=int))) for _ in range(3)]
```

rettigl: I see, this is probably the pythonic way of writing this. I personally find it a bit more difficult to understand, but I can change it.

daviddoji:
LGTM!

rettigl:
> LGTM!

Can you confirm the PR? Then I don't have to overwrite branch protection.

Issue #92, opened 2022-11-25
Move default values into default config


rettigl:

zainsohail04:
Could you please look into this? This would be a great start to a user documentation since I am also not clear as of now what parameters can or do go in the config file and which don't. 
A guideline for the config would be well appreciated too. I can then set it up with doc building.

rettigl:
This depends a little bit on what we consider the default should do/be. I think this is something we should discuss in a meeting.

rettigl:
This is addressed in PR #131

PR #131, opened 2023-06-19
Default config and documentation

rettigl:
This PR moves all default values out of the code and into the default config, and adds respective comments. Stuff from FLASH is still missing (feel free to add).


rettigl:
Closes #92

zainsohail04:
Thanks for the addition.

In the previous PR, you suggested that the defaults should stay within userspace but now you mentioned to add FLASH default config. 

I still can't figure the best way to handle this. Do we add (additional) default values only specific to flash in the default.yaml or create another config file such as flash_defaults.yaml, and then e.g. depending on which loader is being used, it uses the correct default to fill the gaps?

rettigl:
What I mean is that everything should use the config mechanism we set up. The defaults should be in the default config as part of the package and work for our examples, so also for the flash loader, but the user can easily overwrite any given entry by providing an alternative in the config dict to the processor instance without any code change.

zainsohail04:
I have added the flash defaults, what I found necessary.

I am not sure how well the defaults for nexus would translate to flash or other loaders. Or is it somehow general?

We could also try to make the column aliases unified. We have been using camel case for column names but I notice you use snake case. The idea to use camel case was to seperate it from the code variables. But eitherways, since many fields are similar, we could unify.

sed/config/default.yaml
```diff
@@ -2,6 +2,14 @@ core:
   # The loader to use. The generic loader allows for loading various files types directly using dask.
   loader: generic
 
+  # If loader is set to flash, the following parameters are necessary
+  # Needs to be user defined
+  beamtime_id: None
+  year: None
```
rettigl, line 8: You cannot provide None objects in yaml. They end up as "None" strings. In addition, it also does not make any sense to provide None defaults. You anyways catch the case that any of these entries is not present, and then return a meaningful error message. So just leave them away, and have the user define them for a given experiment.

sed/config/default.yaml
```diff
@@ -2,6 +2,14 @@ core:
...
+  # Default beamline is pg2 and instrument is hextof
+  beamline: pg2
+  instrument: hextof
```
rettigl, line 11: Maybe we rename these to flash_beamline and flash_instrument? Then it is clear they belong to the FLASH loader...

sed/config/default.yaml
```diff
@@ -32,6 +40,18 @@ dataframe:
   # binning factor used for the adc coordinate (2^(adc_binning-1))
   adc_binning: 2
 
+  # Flash loader requires the daq parameter and offset
+  daq: fl1user3
+  ubid_offset: 5
```
rettigl, line 45: Explain what they siginify in the comments.

sed/config/default.yaml
```diff
@@ -32,6 +40,18 @@ dataframe:
...
+  # The h5 file paths and aliases are defined under channels
+  # PulseId channel(s) are necessary
+  channels:
+      pulseId:
```
rettigl, line 49: I think this needs to be nested one level less.

sed/config/default.yaml
```diff
@@ -32,6 +40,18 @@ dataframe:
...
+      axis: 1
+      format: per_electron
+      group_name: "/uncategorised/FLASH.EXP/HEXTOF.DAQ/DLD1/"
+      slice: 2
+
```
rettigl, line 54: I realize now we have two different "formats" to address the channels in the hdf5 files: 
For mpes, we have a list of h5 groupnames, and a list of aliases to assign. 
For flash you have this dictionary tree for every channel. We could in prinicple also adopt this for mpes, then it is consistent. I would, however, not provide any channels in the default config, because this cannot be generic for all loaders, I think. Otherwise we shoould split this off into seperate sections. Maybe needs discussion.

rettigl (commented):
I left some comments, I think there are still some bugs with the config you provided, the tests don't pass. 
Regarding the nexus section, the input_files could probably be removed and be part of the specific examples with specific input files for mpes and flash. There, the mapping from metadata to nexus paths is defined.
Regarding dataframe naming: There are two levels here:
-The first are keys that hold the names of the actual dataframe columns: config[dataframe][x_column], etc. These are used in a "pythonic" way and snake_case.
-The second is the actual column names. They are in principle up to the user to define, but I tried to keep them short so far and mostly lower or CamelCase (timeStamps, X, Y, t, delay, energy, etc.). I suggest the defaults we provide here should be as consistent as possible, but I don't see any conflicts so far.

sed/config/default.yaml
daviddoji: Nit pick: That's not entirely true. You can use `None` in yaml files if you add `!!python/none`, e.g.
```E0: !!python/none None```
That's perfectly valid yaml. If that's make sense or not is a different story.

sed/config/default.yaml
zainsohail04: I think a dictionary like structure is more pythonic and the alias just becomes the key of each groupname/path (value).

You are right that it won't be generic anymore so I will remove it.

sed/config/default.yaml
zainsohail04: These are no longer necessary. The initial insight to add them was to seperate the hextof and wespe instrument (and for future), even though they follow the same structure. The only difference being that wespe has two detectors. Since, sed can not support two dataframes simultaneously, for now, we agreed to just run two instances of sed for the two different detectors, and hence two different config files (I made appropriate changes to mark and save different detectors in the loader).

sed/config/default.yaml
zainsohail04: You are right about it. But I just added it here for case of testing. But perhaps creating a dict within the test context is better.

sed/config/default.yaml
zainsohail04: Yes that's possible. I will make an issue out of that, as flash loader needs to adapt to this.
And then mpes can also be adapted to make use of the channels header.

sed/config/default.yaml
rettigl: See comment to flash_reader: Why do you not create one dataframe that has columns for both detectors?

sed/config/default.yaml
rettigl: My point is that the whole thing probably does not work, because pulseId is indented two levels from channels (so it appears to me)...

sed/config/default.yaml
rettigl: We should probably provide a general default config, and examples for each experiment, that can be used to set up a per-site configuration. The mechanism for this is now available in PR #111

rettigl:
Please fix your additions such that the pipeline works again.

sed/config/default.yaml
rettigl: @zainsohail04 I fixed this now, at least tests pass again. Please make sure that you don't break things with your commits in the future, or at least fix them quickly. Here I even told you what to do...

zainsohail04: Hi Laurenz sorry about that. There was still implementation of flash loader that has been changing a lot. Plus, a lot of config handling also changed.
I can revert my commit completely and later add flash defaults once that branch is merged. What do you think?

rettigl:
I reverted all loader-specific entries from the default config, and added loader-specific example configs. Flash config still needs documentation, otherwise this is good to go from my side.

zainsohail04:
> I reverted all loader-specific entries from the default config, and added loader-specific example configs. Flash config still needs documentation, otherwise this is good to go from my side.

I have updated with defaults and comments from flash

zainsohail04 (approved):

PR #133, opened 2023-06-27
support for wespe and save/load parquet

zain-sohail:
- Flash loader now supports reading and saving parquet
- The two detector system of wespe should now be possible using flashloader directly


rettigl:
Fix tests before review, please. For my understanding: What is the second detector of WESPE used for? Why not create a dataframe that contains columns for both detectors, instead of using two instances of the processor?

zainsohail04:
> Fix tests before review, please. For my understanding: What is the second detector of WESPE used for? Why not create a dataframe that contains columns for both detectors, instead of using two instances of the processor?

The two detectors get electrons at different pulses so there is little overlap between them. If we merge the two detectors, firstly there is just a whole lot of NaNs. Secondly, they need to go through an intermediate step of simulations and need to be processed differently/seperately.

Seems like since parquet is now supported, I need to provide a test file for parquet.

zainsohail04:
switch metadata error to warning. also how to concat many metadata runs.

rettigl:
I think there went something wrong with your merge. Maybe it's best to rebase onto main...

zainsohail04:
> I think there went something wrong with your merge. Maybe it's best to rebase onto main...

right that is cleared now.

@rettigl the test assertion here fails https://github.com/OpenCOMPES/sed/blob/f8ec08406a8873e0d6855532b3156c5dfddafb97/tests/loader/test_loaders.py#L147

where  loaded_dataframe.npartitions = 1 and expected_size = 2, because runs has this dict:
`{'generic': None, 'mpes': None, 'flash': ['43878', '43878']}` I have tried to figure this bug out for a while but can't figure it out.

Regarding, adding the parquet in loader supported_file_types, since there is a flag necessary in the loader to load parquet files, it is not finding it. I am thinking to just remove parquet from supported_file_types.

rettigl:
> > I think there went something wrong with your merge. Maybe it's best to rebase onto main...
> 
> right that is cleared now.
> 
> @rettigl the test assertion here fails
> 
> https://github.com/OpenCOMPES/sed/blob/f8ec08406a8873e0d6855532b3156c5dfddafb97/tests/loader/test_loaders.py#L147
> 
> where loaded_dataframe.npartitions = 1 and expected_size = 2, because runs has this dict: `{'generic': None, 'mpes': None, 'flash': ['43878', '43878']}` I have tried to figure this bug out for a while but can't figure it out.
> 
> Regarding, adding the parquet in loader supported_file_types, since there is a flag necessary in the loader to load parquet files, it is not finding it. I am thinking to just remove parquet from supported_file_types.

I'll have a look later

rettigl:
I can tell you why this does not work anymore, but I cannot easily fix it because you have internally changed the way the reader works, thus breaking the compatibility to the intended way the template class functionality should work as I see it. 
The way it is intended to work is the following: 
-If the reader gets either a list of files or folders, it should just take these files, and load them. Period. No raw_data folders or parquet folders or anything involded. The respective test files are expected to be found in the tests/data/reader/flash/ folder. In the case of h5, where conversion is needed, the reader used to use the parquet_folder config entry to know where to place the intermediate files, I suggest to maintain this behavior. But for parquet files, simply load the provided path (list of strings, placed in self.files by the super class method).
-If the reader gets runs, locate them either starting from the "folder" argument, if provided, or from the default, in your case the raw_data_dir. For parquet, I would do the same, then you will also find the runs, if they are placed there. You can also point raw_data_dir and parquet_dir to the same location for the tests, but I don't know how you want to deal with this additional per_file subfolder. Anyways, I don't know why you need this...

rettigl:
> @rettigl the test assertion here fails
> 
> https://github.com/OpenCOMPES/sed/blob/f8ec08406a8873e0d6855532b3156c5dfddafb97/tests/loader/test_loaders.py#L147

It works for the h5 loop case, but fails for the parquet case, because it does not find the test files, for the reason given above. The reasoning with the list above is that I want to test a list with two runs, but as there is only one, I load it twice, and thus expect a dataframe with two partitions, This used to work, if it does not work anymore I don't know why right away now.

rettigl:
> It works for the h5 loop case, but fails for the parquet case, because it does not find the test files, for the reason given above. The reasoning with the list above is that I want to test a list with two runs, but as there is only one, I load it twice, and thus expect a dataframe with two partitions, This used to work, if it does not work anymore I don't know why right away now.

The immediate reason this fails is that you put the filenames from the list into a dict in line 642, thus overwriting the re-occuring filename. 

In general, I don't understand the purpose of most of the changes you are implementing here in this PR, so I can neither comment nor review.

rettigl:
A lot of incomplete/wrong docstrings and missing typing information is also still existing...

zainsohail04:
> > @rettigl the test assertion here fails
> > https://github.com/OpenCOMPES/sed/blob/f8ec08406a8873e0d6855532b3156c5dfddafb97/tests/loader/test_loaders.py#L147
> 
> It works for the h5 loop case, but fails for the parquet case, because it does not find the test files, for the reason given above. The reasoning with the list above is that I want to test a list with two runs, but as there is only one, I load it twice, and thus expect a dataframe with two partitions, This used to work, if it does not work anymore I don't know why right away now.

For me, it was failing for h5 case as well. I removed parquet and tried to test but was still giving the list of 2.



> A lot of incomplete/wrong docstrings and missing typing information is also still existing...

I believe that shouldn't be the case? I made sure to document the loader extensively but perhaps I missed somewhere.

rettigl:
> For me, it was failing for h5 case as well. I removed parquet and tried to test but was still giving the list of 2.

You are right, the reason is the one I wrote in my reply afterwards

zainsohail04:
>>  The immediate reason this fails is that you put the filenames from the list into a dict in line 642, thus overwriting the re-occuring filename.

I think we had a discussion about this before. It's part of the loaders job to figure out the file structure, as it is quite beamtime/instrument dependant and requires the config parameters to figure out. Hence we create filenames from each run.



>> -If the reader gets either a list of files or folders
we never use this situation, hence why we got the runs included

>> If the reader gets runs, locate them either starting from the "folder" argument, if provided, or from the default, in your case the raw_data_dir. For parquet, I would do the same, then you will also find the runs, if they are placed there. You can also point raw_data_dir and parquet_dir to the same location for the tests, but I don't know how you want to deal with this additional per_file subfolder. Anyways, I don't know why you need this...

The thing is, we don't use those two parameters generally. Especially on maxwell.

>> In general, I don't understand the purpose of most of the changes you are implementing here in this PR, so I can neither comment nor review.

Mostly to allow the wespe instrument to function, which requires that we be able to save/load parquets.

The intermediate parquets are just for faster access when reloading data.

sed/loader/flash/loader.py
```diff
@@ -41,11 +44,129 @@ class FlashLoader(BaseLoader):
...
+                    ):
+                        data_raw_dir.append(path.joinpath(daq))
+                    elif dir_name == daq.upper():
+                        data_raw_dir.append(path)
+
+            if not data_raw_dir:
+                raise FileNotFoundError("Raw data directories not found.")
+
+            parquet_path = "processed/parquet"
+            data_parquet_dir = beamtime_dir.joinpath(parquet_path)
```
rettigl, line 106: This whole branch is not being tested for. I think it would be good to write a separate test for it, but could be done later.

sed/loader/flash/loader.py
```diff
@@ -496,14 +609,72 @@ def h5_to_parquet(self, h5_path: Path, parquet_path: Path) -> None:
...
+                .reset_index(level=self.multi_index)
                 .to_parquet(parquet_path, index=False)
             )
         except ValueError as failed_string_error:
             self.failed_files_error.append(
                 f"{parquet_path}: {failed_string_error}",
             )
-            self.parquet_names.remove(parquet_path)
+
+    def buffer_file_handler(self, data_parquet_dir, detector):
```
rettigl, line 620: Add typing?

sed/loader/flash/loader.py
```diff
@@ -561,6 +730,76 @@ def fill_na(
         # Concatenate the filled dataframes
         return dd.concat(dataframes)
 
+    def parquet_handler(self, data_parquet_dir, **kwds):
```
rettigl, line 733: I would suggest to handle the parameters you list the same way we do in all other functions, i.e. list them as optional parameters with typing information = None, and handle them in the function accodingly, rather than using kwds as you do now. Or is there a strong reason doing this? Right now, there is no type checking done, and it is intransparent which parameters the function takes from the function prototype.

sed/loader/flash/loader.py
```diff
@@ -561,6 +730,76 @@ def fill_na(
...
+        Raises:
+            FileNotFoundError: If the requested parquet file is not found.
+
+        """
+
+        # Creates the identifier for detector, if the parameter exists
+        detector = f"_{kwds.get('detector')}" if "detector" in kwds else ""
+
+        # Get the parquet path from the keyword arguments, or construct it if not provided
+        parquet_path = kwds.get("parquet_path")
```
rettigl, line 757: If you introduce this as a parameter, you just can delete this line...

tests/data/loader/flash/parquet/run_43878.parquet
rettigl, line 1: remove file, as no longer needed

sed/loader/flash/loader.py
```diff
@@ -561,6 +730,76 @@ def fill_na(
...
+        if parquet_path is None:
+            parquet_name = "_".join(str(run) for run in self.runs)
+            parquet_dir = (
+                data_parquet_dir.joinpath("converted")
+                if kwds.get("converted", False)
```
rettigl, line 762: converted is not mentioned as parameter

sed/loader/flash/metadata.py
rettigl, line 1: There are no tests for metadata yet. I think mpes metadata are tested partly through the processor tests, maybe you can do something similar, or add a generic metadata test in test_loader?

rettigl: Could also be an issue

rettigl (approved):
I left a few remarks where things can still be improved, otherwise LGTM

sed/loader/flash/loader.py
zain-sohail: Yes, soon we have some data publicly accessible so I can test this with much more ease than would be possible elsewise.

sed/loader/flash/metadata.py
zain-sohail: referenced in #154

sed/loader/flash/loader.py
zain-sohail: refereced in #154

PR #136, opened 2023-07-30
Mpes tweaks

rettigl:
Some further improvements for the mpes workflow an loader


rettigl:
This PR contains a few usefull additions mostly for mpes:
- print some output from the correction workflows into widgets, as normal output is not shown in jupyter lab after a cell has been executed
- revert the momentum correction back before successive application of pose corrections
- add "runs" implementation for the mpes loader
- add option to pass a eln file to the nexus converter, without providing the complete file list

I want to merge it before it grows any bigger yet, a review should be still quick and feasible.

pyproject.toml
```diff
@@ -31,6 +31,7 @@ tqdm = "^4.62.3"
 xarray = "^0.20.2"
 joblib = "^1.2.0"
 jupyter = {version = "^1.0.0", extras = ["notebook"], optional = true}
+jupyterlab = {version = ">=3.2.9, <4.0.0", extras = ["notebook"], optional = true}
```
steinnymir, line 34: do newer versions of jupyter lab give problems? What is incompatible>

steinnymir (approved):
Looks like mostly MPES related changes, so nothing against that.
Only wondering why limiting the jupyter version.

pyproject.toml
rettigl: In versions above 4.0, for some reason widgets seem to not properly output anymore. I don't have any idea why, and websearches don't give recent hits. Even graphs with "%matplotlib widget" don't show, but give something like this:
grafik
Feel free to test yourself, the pinning is a rather dirty fix...

daviddoji: Make sure that the versions of 'ipympl', 'matplotlib' and 'ipywidgets' from Jupyter and the one you use in your environment do match. That should fix the problem

rettigl: Thanks, indeed the limit on ipywidget was te problem. I removed this, now it works.

Issue #146, opened 2023-09-09
Add loading of energy calibration traces from arrays to processor interface


rettigl:
ad possibility to load energy calibration traces into the energy calibrator class from the processor interface from already binned data, without invoking binning function

rettigl:
Addressed in PR #152

rettigl:
This code, however, still needs tests, once the processor_tests branch is merged!

Issue #102, opened 2023-02-07
Energy calibration for flash


yacremann:
For operation at FLASH we developed a calibration method of the energy calibration (it is part of the hextof processor). It would be great to have the same functionality in SED.

rettigl:
The calibration fitting functions presently there should also work for FLASH bias series. You can also directly supply the calibration parameters to the function. How did you perform energy calibration at FLASH before? I would prefer if we find a common solution, so that we don't have multiple functions for doing similar things.

steinnymir:
We are currently working on implementing the corrections and calibrations from `hextof-processor`, but are currently still not available.

The principle behind the energy calibration there was to transform the time-of-flight to an energy scale, based on photon energy, sample bias and tof bias, all from measured quantities. Optionally these could be also fixed values.

This has the advantage of correcting for possible fluctuations in the photon energy, but also has some drawbacks. In particular, the electron path is assumed constant for all energies and detected angles, which is a rough approximation.

I believe we should implement at first all the hextof and all the MPES methods and make them easily callable. Then we can select which to prioritize or try to combine them when possible.

rettigl:
I think you should be able to get a similar result by adjusting the offset parameter accordingly. But off course you can add another method to the function as well. I would just like to avoid that we have multiple functions and/or modules that do more or less the same. This will confuse users.

steinnymir:
I agree that we should try avoid multiple functions doing the same thing, but also loosing functionality from other methods because one is already implemented is not ideal.

having two different approaches that to do similar things, does not make them interchangeable. There are some corrections necessary for FLASH data which do not fit in the workflow for MPES (sector correction for example).

I believe a good starting point is to provide different workflow methods which can be applied at will on a workflow manager. This can then be tuned to use the preferred methods for each experiment.
I will explain this in more detail in our meeting Tuesday.

zainsohail04:
Hi @steinnymir, any updates on the workflow feature? Dima was saying we need the calibration functions soon so I was wondering if we should try implementing them in the current state or with the workflow feature.

steinnymir:
hi @zainsohail04, I am still at an early stage of the infrastructure, but I think I could make the pull request soon. There will still be many things to implement to get a full, flexible workflow though.

When is the beamtime starting?

steinnymir:
a first basic version is in the pull request #104

kutnyakhov:
[...]

PR #152, opened 2023-09-14
Flash energy calibration

rettigl:
This PR is for developing energy calibration methods for flash. 
So far, it contains an example notebook on how to load the provided FLASH energy calibration run into the existing energy calibration workflow in the processor, and demonstrates how to use the existing energy calibration functions.

### DIFF
--- a/sed/config/default.yaml
+++ b/sed/config/default.yaml
@@ -11,6 +11,8 @@ dataframe:
   tof_column: "t"
   # dataframe column containing analog-to-digital data
   adc_column: "ADC"
+  # dataframe column containing bias voltage data
+  bias_column: "sampleBias"
   # dataframe column containing corrected x coordinates
   corrected_x_column: "Xm"
   # dataframe column containing corrected y coordinates
--- a/sed/core/processor.py
+++ b/sed/core/processor.py
@@ -834,7 +834,8 @@ class SedProcessor:
     # 1. Load and normalize data
     def load_bias_series(
         self,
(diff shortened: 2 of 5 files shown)


steinnymir:
What are the changes introduced is this PR?
should I merge them into #169 where we are developing hextof related calibrations and corrections?

rettigl:
It contains the energy calibration notebook by dima, which I modified to use the methods from mpes to optain the parameters, and do the energy calibration. For this I had to also modify the code somewhat to allow ingesting the bias series data from externally binned arrays, rather than only binning internally. This was on the agenda anyways.
The notebook could either stay a tutorial notebook, or be removed, otherwise this could be merged to main, as it closes #146.
It also provides a first demonstrator for #102, albeit without the energy shift by bias/photon energy, etc, and potential changes you might want to introduce.

rettigl:
It might not be a good tutorial notebook, as the data are not publically available. We could upate it with the now-available public data.

steinnymir:
ok, so these changes will be necessary to run the energy calibration functions.
I would merge this in #169 then. is this tested and ready to go?

rettigl:
Apart from the notebook, this is has tests and works. It's a small change. As I said, after review can be merged to main to close #146 and then rebase #169, that would be my preferred way

steinnymir (approved):

steinnymir:
Can we merge this PR so I can implement it in #169 ?

PR #162, opened 2023-10-07
simplify NaN filling in FlashLoader and make it lazy

zain-sohail:
After a few attempts with different options, I found that apparently (perhaps recently) multiple files can be loaded directly using read_parquet from dask so if I do that and just perform an ffill on the necessary channels, it works. dask seems to be handling the chunking etc. at least from the tests I did

### DIFF
--- a/sed/loader/flash/loader.py
+++ b/sed/loader/flash/loader.py
@@ -737,8 +737,13 @@ class FlashLoader(BaseLoader):
             # Channels to fill NaN values
             channels: List[str] = self.get_channels_by_format(["per_pulse", "per_train"])
 
-            # Fill NaN values
-            dataframe[channels] = dataframe[channels].ffill()
+            # Define a custom function to forward fill specified columns
+            def forward_fill_partition(df):
+                df[channels] = df[channels].ffill()
+                return df
+
+            # Use map_overlap to apply forward_fill_partition
+            dataframe = dataframe.map_overlap(forward_fill_partition, before=0, after=1)
 
             # Remove the NaNs from per_electron channels
             dataframe = dataframe.dropna(
--- a/sed/loader/flash/loader.py
+++ b/sed/loader/flash/loader.py
@@ -8,7 +8,6 @@ This can then be saved as a parquet for out-of-sed processing and reread back to
 sed funtionality.
 """
 from functools import reduce
-from itertools import compress
 from pathlib import Path
 from typing import List
 from typing import Sequence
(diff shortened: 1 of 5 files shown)


rettigl:
I thinl that always worked, no? As far as I understand the issue, the problem addressed by fill_nan is that some of the slow channels might not have any valid value within a given parquet file. Then, ffill will not be able to fill in any reasonable value. The solution is to get a value from the previous file, where such a value existed.
This mechanism does not work with this suggested solution, as far as I can tell.

One option might be to split the work of fill_nan up: get a valid value for all slow parameters before storing files, and save them as zeroth event value. Thus we make sure, that later this ffill  approach will always work.

zain-sohail:
> I thinl that always worked, no? As far as I understand the issue, the problem addressed by fill_nan is that some of the slow channels might not have any valid value within a given parquet file. Then, ffill will not be able to fill in any reasonable value. The solution is to get a value from the previous file, where such a value existed. This mechanism does not work with this suggested solution, as far as I can tell.
> 
> One option might be to split the work of fill_nan up: get a valid value for all slow parameters before storing files, and save them as zeroth event value. Thus we make sure, that later this ffill approach will always work.

You are right. That was not the complete solution. Actually I wrote the fill nan method long before.
But anyways, I came up with another idea (using map_overlap on dataframe) to do it lazy but just missed commiting it

rettigl:
> > I thinl that always worked, no? As far as I understand the issue, the problem addressed by fill_nan is that some of the slow channels might not have any valid value within a given parquet file. Then, ffill will not be able to fill in any reasonable value. The solution is to get a value from the previous file, where such a value existed. This mechanism does not work with this suggested solution, as far as I can tell.
> > One option might be to split the work of fill_nan up: get a valid value for all slow parameters before storing files, and save them as zeroth event value. Thus we make sure, that later this ffill approach will always work.
> 
> You are right. That was not the complete solution. Actually I wrote the fill nan method long before. But anyways, I came up with another idea (using map_overlap on dataframe) to do it lazy but just missed commiting it

The idea is good, but not sure your implementation does the right thing yet. Looking at the doc of map_overlap, the before, after refer to single rows, I. E. Entries. Might be it picks a filled one of a slow column, might be not. Anyways, you want to use before, no? Do we have test data that actually allow to test this problem?

zain-sohail:
> > > I thinl that always worked, no? As far as I understand the issue, the problem addressed by fill_nan is that some of the slow channels might not have any valid value within a given parquet file. Then, ffill will not be able to fill in any reasonable value. The solution is to get a value from the previous file, where such a value existed. This mechanism does not work with this suggested solution, as far as I can tell.
> > > One option might be to split the work of fill_nan up: get a valid value for all slow parameters before storing files, and save them as zeroth event value. Thus we make sure, that later this ffill approach will always work.
> > 
> > 
> > You are right. That was not the complete solution. Actually I wrote the fill nan method long before. But anyways, I came up with another idea (using map_overlap on dataframe) to do it lazy but just missed commiting it
> 
> The idea is good, but not sure your implementation does the right thing yet. Looking at the doc of map_overlap, the before, after refer to single rows, I. E. Entries. Might be it picks a filled one of a slow column, might be not. Anyways, you want to use before, no? Do we have test data that actually allow to test this problem?

You're right about picking or not, depends on whether there is an entry at the end or not. I have been having a deeper look at it. 
I think the only way to support an empty or series of empty chunks is to do this operation in a serial manner, unless we have apriori information that there won't be an empty chunk in a certain size of data (asking dima is a good start)

steinnymir:
I changed the parameters of the map_overlap function to load n lines from the previous partition with `n = min(partition sizes)`.
This seems to work correctly now and adds just a few seconds to the loading time (compared to 10 minutes with the previous fill_na)

I also merged this change in my current work in progress branc WorkflowManager2

steinnymir (approved):
With the correction I implemented, I think this is good to go.
Havent checked tests though. maybe we could implement some proper tests for this too.

rettigl:
> I also merged this change in my current work in progress branc WorkflowManager2

No, you merged your changes into this PR. I don't think this is a good idea...

sed/loader/flash/loader.py
```diff
@@ -786,11 +736,27 @@ def parquet_handler(
...
+
+            # Define a custom function to forward fill specified columns
+            def forward_fill_partition(df):
+                df[channels] = df[channels].ffill()
+                return df
+
+            # calculate the number of rows in each partition
+            with ProgressBar():
+                print("Computing dataframe shape...")
+                nrows = dataframe.map_partitions(len).compute()
```
zain-sohail, line 753: These values should be in the parquet metadata called [num_rows](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.FileMetaData.html#pyarrow.parquet.FileMetaData.num_rows) [statistics](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.Statistics.html#pyarrow.parquet.Statistics) generally.
statistics, which dask also loads, have metadata like null_count etc
https://github.com/dask/dask/blob/928a95aa56f60da33a4e724ea2ca97797c612968/dask/dataframe/io/parquet/core.py#L569
And hence that's what I am trying to figure out: how to access those null_counts and that would make it very easy to know if we have an nan only column or not.
Your idea is also very good. If we can just directly use the metadata num_rows, there would be no need for compute.

steinnymir: If you find access to these informations that would be great. I always found it weird it needs to compute this... but I guess there are lazy operations which could delete rows, making it a not really fixed value.

for now, all my tests returned 0 nans, and computing the lenght takes a few secs, but if that goes, even better!

zain-sohail:
closing as new PR created #165

steinnymir:
Ops, I did not mean to merge the workflow related changes here. My bad! thanks for taking care of this @zain-sohail

PR #165, opened 2023-10-09
NaN filing between partitions with overlap

zain-sohail:
Created a new branch because the one in PR #162 has too many other comments. I think after adding a test, this is good to go.

Changes from the other PR:
- removed progress bar as it has a larger overhead than just calling len
- removed the compute command because len directly gives the metadata access to the num_rows.

To be done:

adding tests with multiple files
- [x] partial data
- [x] one fully nan file (cols)
- [x] consective empty files 

last one might fail so we would need null metadata for a more comprehensive treatment https://github.com/OpenCOMPES/sed/pull/162#discussion_r1350242308

also might make sense to keep this as a seperate function or even a generic method that other loaders can use as I think it's pretty useful


sed/loader/flash/loader.py
```diff
@@ -786,11 +731,29 @@ def parquet_handler(
...
+            channels: List[str] = self.get_channels_by_format(["per_pulse", "per_train"])
+
+            # Define a custom function to forward fill specified columns
+            def forward_fill_partition(df):
+                df[channels] = df[channels].ffill()
+                return df
+
+            # calculate the number of rows in each partition and choose least
+            nrows = dataframe.map_partitions(len)
+            max_part_size = min(nrows)
```
steinnymir, line 747: This is here because we cannot prepend data from two partitions away. Therefore the only single number which always works is the minimum size of all partitions.
Maybe there is a way to make map_overlap use always the size of the previous/next partition, but I am afraid not.

sed/loader/flash/loader.py
```diff
@@ -786,11 +731,29 @@ def parquet_handler(
...
+
+            # Use map_overlap to apply forward_fill_partition
+            dataframe = dataframe.map_overlap(
+                forward_fill_partition,
+                before=max_part_size + 1,
```
steinnymir, line 752: the +1 would make this fail, as described in the comment before

zain-sohail: sorry i didn't quite get what you meant in your previous comment. 
isn't this what you pushed in your previous commit? or are you referring to the PR description

steinnymir: I am working on this currently, and will push a solution with tests soon

steinnymir:
I am working on this currently, and found that the map_partition(len) method is indeed lazy, but if you want to get the number out, i.e. min(df.map_partition(len)) this needs to be computed, and takes as long (or longer) than the previously proposed method. did you test this?

steinnymir:
image
I tested two cases, pre-computing the shape (as I developed previously, with progress bar) and the version you included here, which lets min(...) compute. 

The latter is clearly not lazy, and less optimized this way. so I suggest keeping the pre-computed approach. Both are available with a toggle in the dfops funcion I made.

steinnymir:
The test fails, but when running it on a notebook shows no problems. I do not understand the origin of this test failure, as I cannot reproduce it.

steinnymir:
A further addition implements the option to run ffill multiple times, allowing to fill neighboring partitions full of nans.

steinnymir:
your solution is excellent!
faster and allows gathering even more metadata. Can we keep that metadata information somewhere, so we can reuse it later? as a property of the dataframe?

zain-sohail:
> your solution is excellent! faster and allows gathering even more metadata. Can we keep that metadata information somewhere, so we can reuse it later? as a property of the dataframe?

Surely. Maybe not the correct PR for that. We can fix the tests here and then open another branch for proper metadata treatment.

steinnymir:
ok, but lets keep the compute versions for testing and ensuring this works as expected.

steinnymir (approved):

zain-sohail:
> ok, but lets keep the compute versions for testing and ensuring this works as expected.

Yes sure. I think it's good to keep it generic so that other loaders can use it if needed. 

It seems pyarrow is not part of our package but fastparquet is. Any inherent disadvantage if we switch or keep both? 
fastparquet doesn't allow metadata access from what i know

steinnymir:
fastparquet is giving me big headakes at the moment, with version 0.8.3 (the one that gets installed with poetry) giving errors on read_parquet.
It might be an option to look into changing everything to pyarrow, but in any case I have nothing against having them both as requirement.

steinnymir:
> > A further addition implements the option to run ffill multiple times, allowing to fill neighboring partitions full of nans.
> 
> I like this addition overall. A seperate dfops function that others can use is also great.
> 
> > The test fails, but when running it on a notebook shows no problems. I do not understand the origin of this test failure, as I cannot reproduce it.
> 
> It seems it has to do with a generic test written for all loaders. Somehow multiple dataframes are being returned? But anyways running the tests locally always gives better insight.
> 
> > I am working on this currently, and found that the map_partition(len) method is indeed lazy, but if you want to get the number out, i.e. min(df.map_partition(len)) this needs to be computed, and takes as long (or longer) than the previously proposed method. did you test this?
> 
> Hi. You are probably right with that. I was not trying to do it lazy but to just directly access the metadata, which takes no computation. But seems like dask is not giving access to that.
> 
> I have come up with another solution to it. Basically to use pyarrow to read the metadata for each file. Upper one is my function and lower is the one which lazy computes the shape. So it seems, even with the read operation, it is faster to just load metadata associated with the parquets. Screenshot 2023-10-10 at 16 17 54
> 
> This also allows us to access other metadata which could in principle allow us to do a better treatment. Screenshot 2023-10-10 at 16 28 25 Hence I will push this change and we can build on that

this even seems to show how many nans we have on a column! would allow to programmatically choose how many fillna are needed for each channel!

PR #166, opened 2023-10-09
Workflow restructuring

rettigl:
As discussed:
- run all-versions tests only on push to main (pull-request merges)
- speed up tests by omitting multiple calculations of inverse dfields in metadata tests


rettigl:
Disadvantage is: We apparently don't get the coveralls reports if we don't do pull-request builds...

zain-sohail:
Looks good. Definitely faster now.
I will look into caching the installation of the package as that is mostly same. Would save another 2 mins. 
And you said the other slowdown was downloading data for tests right? Maybe we can also cache that, or use a smaller set?

rettigl:
> Looks good. Definitely faster now. I will look into caching the installation of the package as that is mostly same. Would save another 2 mins. And you said the other slowdown was downloading data for tests right? Maybe we can also cache that, or use a smaller set?

That was related to the notebook conversion. I suppose we don't do this often.

zain-sohail:
> > Looks good. Definitely faster now. I will look into caching the installation of the package as that is mostly same. Would save another 2 mins. And you said the other slowdown was downloading data for tests right? Maybe we can also cache that, or use a smaller set?
> 
> That was related to the notebook conversion. I suppose we don't do this often.

Right. I am not sure if the github workers can actually process the notebooks. I will try that and see. Otherwise, we would have to do it locally or our own runner (possible to set this up now). 

I will merge this PR and update a few workflows afterwards.

zain-sohail (approved):

PR #168, opened 2023-10-09
add default jitter axes to default config, and add tests for that. 

rettigl:
Also, add **kwds arguments to add_jitter

### DIFF
--- a/sed/config/default.yaml
+++ b/sed/config/default.yaml
@@ -31,6 +31,8 @@ dataframe:
   tof_binning: 1
   # binning factor used for the adc coordinate (2^(adc_binning-1))
   adc_binning: 1
+    # list of columns to apply jitter to
+  jitter_cols: ["@x_column", "@y_column", "@tof_column"]
 
 energy:
   # Number of bins to use for energy calibration traces
--- a/sed/config/default.yaml
+++ b/sed/config/default.yaml
@@ -31,8 +31,10 @@ dataframe:
   tof_binning: 1
   # binning factor used for the adc coordinate (2^(adc_binning-1))
   adc_binning: 1
-    # list of columns to apply jitter to
+  # list of columns to apply jitter to.
   jitter_cols: ["@x_column", "@y_column", "@tof_column"]
+  # Jitter amplitude or list of jitter amplitudes. Should equal half the digitial step size of each jitter_column
+  jitter_amps: [0.5, 0.5, 0.5]
 
 energy:
   # Number of bins to use for energy calibration traces
--- a/sed/core/processor.py
+++ b/sed/core/processor.py
@@ -1180,23 +1180,25 @@ class SedProcessor:
             else:
                 print(self._dataframe)
 
-    def add_jitter(self, cols: Sequence[str] = None):
+    def add_jitter(self, cols: List[str] = None, **kwds):
         """Add jitter to the selected dataframe columns.
 
         Args:
-            cols (Sequence[str], optional): The colums onto which to apply jitter.
+            cols (List[str], optional): The colums onto which to apply jitter.
                 Defaults to config["dataframe"]["jitter_cols"].
+            **kwds: keyword arguments passed to apply_jitter
         """
         if cols is None:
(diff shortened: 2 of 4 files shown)


sed/core/processor.py
```diff
@@ -1180,23 +1180,25 @@ def calibrate_delay_axis(
             else:
                 print(self._dataframe)
 
-    def add_jitter(self, cols: Sequence[str] = None):
+    def add_jitter(self, cols: List[str] = None, **kwds):
```
steinnymir, line 1183: why use a different behaviour for cols and amps? should we not add it as argument here with default to None?

sed/config/default.yaml
```diff
@@ -31,6 +31,8 @@ dataframe:
   tof_binning: 1
   # binning factor used for the adc coordinate (2^(adc_binning-1))
   adc_binning: 1
+    # list of columns to apply jitter to
+  jitter_cols: ["@x_column", "@y_column", "@tof_column"]
```
steinnymir, line 36: I would suggest addding here `jitter_amps: [0.5,0.5,0.5]`. this would be redundant here, but in the case of hextof would actually be `jitter_amps: [0.5,0.5,4]`

steinnymir (changes requested):
I would even consider changing the default value of the apply_jitter attribute amp to be 1 instead of 0.5, and half it when we define the jitter values, as `-amp/2, amp/2` instead. What do you think? this would make the amps in the config file cleaner.

rettigl:
> I would even consider changing the default value of the apply_jitter attribute amp to be 1 instead of 0.5, and half it when we define the jitter values, as `-amp/2, amp/2` instead. What do you think? this would make the amps in the config file cleaner.

Hmm, we can discuss this, I don't have a strong preference. What is easier to understand?

sed/core/processor.py
rettigl: I thought that amps should not be a default argument, because we would not per default use it. I wanted to provide access to the other arguements of the function, in case they are needed, but they should not be needed per default, as jittering should only be required on digital values, and then the jittering should be done with amp=0.5

sed/config/default.yaml
rettigl: My very strong preference would be to fix this in the flash loader, and have correct digital units on the flash dldTime, and leave amps out of the config... we still have this kwd access in case one needs it for debugging.

steinnymir:
I feel the full width of the jitter size (i.e. steps of 1) is more intuitive, as we add jitter with the amplitude of 1 bin. The other case feel like "we add jitter with amplitude twice a half bin".

sed/config/default.yaml
steinnymir: I disagree though, as this is not general. 
There might be other cases in which a digial channel is not represented in units of the digital step.
We don't have an example here, but I had similar experiences with continuous values (voltage) measured through an ADC which had floating point step size. Was digital but represented as float.

And why such a strong preference? does this break something or make something hard to understand?

steinnymir (approved):

sed/config/default.yaml
steinnymir: I fixed this in the flash loader

rettigl: > And why such a strong preference? does this break something or make something hard to understand?

Having exactly the right jitter parameters is quite important if you want the digitzation artefacts to disappear, but no loss of resolution. That's why I would rather keep this away from the user and as the responsibility of the instrument scientist setting up the config file to check this carefully. We should add an example notebook for the influence of the jittering.

rettigl:
> I feel the full width of the jitter size (i.e. steps of 1) is more intuitive, as we add jitter with the amplitude of 1 bin. The other case feel like "we add jitter with amplitude twice a half bin".

The code applies "amp*uniform", that's why I think keeping the current definition is the more correct one. I added a better description.

sed/config/default.yaml
rettigl: Having said this, I does not provide a very good reason against having it as a config parameter, so I have added it now.

PR #174, opened 2023-10-11
add rolling average and energy shift

steinnymir:
add the ability to run a rolled average on a column, based on acquisition time instead of on the direct rows. Useful for smoothing noisy channels.

### DIFF
--- a/sed/calibrator/hextof.py
+++ b/sed/calibrator/hextof.py
@@ -10,6 +10,7 @@ from typing import Union
 import numpy as np
 import pandas as pd
 import dask.dataframe
+from dask.diagnostics import ProgressBar
 
 
 def unravel_8s_detector_time_channel(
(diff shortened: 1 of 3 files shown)


sed/calibrator/hextof.py
rettigl, line 1: Can your break this file up, and put the repsective functions into the calibrator files related to the physical axes? "hextof" is not a physical axis.
So, put energy functions into energy, momentum into momentum, etc. and for the tof-related stuff (dld calibration) make another file tof.py.
I think there should not be anything specific in the main code to "mpes" or "hextof" or so, rather than the loaders, and that's how I tried to design the rest of the code. Off course one or the other function will only work in specific contexts, but those should not be limited to specific instruments.

steinnymir: I dont intend to merge the hextof folder. I just find the thousands of lines of those files scary! 😓 

I'll trasnfer everything to its proper place, promised!

PR #178, opened 2023-10-13
add depedabot for dependency tracking/updating

zain-sohail:
With the discussion in PR #151 , we want automatic dependency updates every so often. Depedanbot is a tool by github to perform just this function. It creates a PR weekly (or any time duration) with all the outdated dependencies to update them, and those can be manually merged (or auto if wished). 
Initally I had other ideas on how to do this but dependabot is very well integrated in github environement (see insights or security)

example is PR #179 in which dependabot suggested us a security update for numpy. For general updates, it will only do so once we merge this branch to main and then it will group all the updates in one PR (hard to test without trying).


rettigl:
Well, I think there were good reasons we fixed at least some of the package versions to where they are. I tried to solve the mypy issues with the new numpy version, but couldn't, at least some of them. For, me, not worth spending the time right now. Updating packages, please automatically, dependencies, only manually.

rettigl:
As long as such PRs don't automatically come up every week again once we decide against updating, I'm also fine with that. In general, it's probably a good idea installing such a guard.

zain-sohail:
> As long as such PRs don't automatically come up every week again once we decide against updating, I'm also fine with that. In general, it's probably a good idea installing such a guard.

The alerts can be dismissed here:
https://github.com/OpenCOMPES/sed/security/dependabot
Screenshot 2023-10-16 at 12 50 42

I will wait for PR approval if there are no additional concerns.

rettigl (commented):
Does this do what we need, i.e. update all package dependencies to the latest versions compliant with our dependency rules? Will this also update the versions in the lock file if there are _no_ security problems, but just a version update? 
What we need, is essentially a
poetry update
comitt + push poetry.lock
in the poetry environment with all options installed

rettigl:
> Does this do what we need, i.e. update all package dependencies to the latest versions compliant with our dependency rules?

Most certainly not. In the np-to-1.22.0 PR, Numpy was set to 1.22.0 in the lock file, whereas it will become 1.22.4 from the pyproject.yaml file if you install via pip, or do poetry update. So no, we need something else in addition.

zain-sohail:
> Does this do what we need, i.e. update all package dependencies to the latest versions compliant with our dependency rules? Will this also update the versions in the lock file if there are _no_ security problems, but just a version update? What we need, is essentially a poetry update comitt + push poetry.lock in the poetry environment with all options installed

What happened in #179 is not the purpose of this PR. The security updates come seperately from dependabot because I enabled that. 
This PR would do the regular weekly updates to the latest versions. Again, I can not be sure but this is what the documentation and my search on the internet has told me.

If it doesn't, it's quite trivial to update poetry lock with a cron job since we already have the infrastructure (workflows) setup.

rettigl (approved):
Maybe you are right and this does what we want. I think it will only run if it is in main, so let's test out.

PR #181, opened 2023-10-14
define jitter_amps as single amplitude in default config

rettigl:


rettigl:
With such a default, one does not need to redefine it for different number of jitter cols in other configs

PR #182, opened 2023-10-16
Improve workflows to use github caches

zain-sohail:
Generally addition of using caches in workflows
Parallization of testing of package and fixes for race conditions
Addition of dev dependencies so linting can be done

Refer to #151 for more details

Overall, workflows are much faster now
Linting: 1m
Test: 3-4m (after caching)


rettigl:
Coverage calculation does not work

.github/workflows/testing_multiversion.yml
```diff
@@ -0,0 +1,51 @@
...
+    needs: pytest
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    steps:
+    - name: Coveralls Finished
+      uses: coverallsapp/github-action@v2
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        parallel-finished: true
+        carryforward: "run-3.8,run-3.9,run-3.10"
```
rettigl, line 51: Did you test this?

rettigl (commented):
In general, I think this is a nice addition, however I have a few concerns/suggestsions:
1. We need to make sure that the poetry.lock file is regularly updated to the latest dependency versions if we want to use this. Dependabot does _not_ do this in its current config (see #178)
2. There is currently no pytesting done on push. This should be added, at least for one python version, for pushes to branches without a PR.
3. Coverage reports do not seem to report correct numbers

rettigl:
The pytest workflow does not even run when pushing to this branch/PR. Seems completely disabled right now.

daviddoji:
Related to this PR, if you also want to speed up on linting, have a look at ruff:
https://docs.astral.sh/ruff/

rettigl:
> @zain-sohail Related to this PR, if you also want to speed up on linting, have a look at ruff: https://docs.astral.sh/ruff/

Interesting, but it does not seem to replace pylint or mypy, which take the major time in our lining pipeline.
Pylint, however, can still be parallelized, -j parameter

zain-sohail:
> > @zain-sohail Related to this PR, if you also want to speed up on linting, have a look at ruff: https://docs.astral.sh/ruff/
> 
> Interesting, but it does not seem to replace pylint or mypy, which take the major time in our linting pipeline. Pylint, however, can still be parallelized, -j parameter

I have thought about the Ruff option. If we can implement a similar rule set (or better) with Ruff, it's definitely faster. But we would still need a type checker like mypy.

.github/workflows/testing_multiversion.yml
zain-sohail: Yes. Here you can see: https://github.com/OpenCOMPES/sed/actions/runs/6510731696

zain-sohail:
> In general, I think this is a nice addition, however I have a few concerns/suggestsions:
> 
> 1. We need to make sure that the poetry.lock file is regularly updated to the latest dependency versions if we want to use this. Dependabot does _not_ do this in its current config (see [add depedabot for dependency tracking/updating #178](https://github.com/OpenCOMPES/sed/pull/178))

Addressed in that PR. We can implement this, if necessary, later.
> 2. There is currently no pytesting done on push. This should be added, at least for one python version, for pushes to branches without a PR.

But this is the same as before. We were never using push. See https://github.com/OpenCOMPES/sed/blob/19921fc33cfb4c491a490e547c61a100badd8128/.github/workflows/pytest_coveralls.yml#L6
I just reimplemented the ruleset.

> 3. Coverage reports do not seem to report correct numbers

Yes, on that I am not quite certain why. Even for single version test, it doesn't. Or the older ones were wrong? Because now I am using the official action from coveralls.io
I will test the coverage locally and see.

daviddoji:
> > @zain-sohail Related to this PR, if you also want to speed up on linting, have a look at ruff: https://docs.astral.sh/ruff/
> 
> Interesting, but it does not seem to replace pylint or mypy, which take the major time in our linting pipeline. Pylint, however, can still be parallelized, -j parameter

Ruff and mypy are tools for different purposes, while the former is a linter, the latter is a type checker. Each of them can detect problems that the other would miss. It's the combination of them what it makes sense to use.

Regarding pylint, maybe it's worth to give it a shot as many users have succeeded (https://docs.astral.sh/ruff/faq/#how-does-ruff-compare-to-pylint)

rettigl:
> > 3\. There is currently no pytesting done on push. This should be added, at least for one python version, for pushes to branches without a PR.
> 
> But this is the same as before. We were never using push. See
> 
> https://github.com/OpenCOMPES/sed/blob/19921fc33cfb4c491a490e547c61a100badd8128/.github/workflows/pytest_coveralls.yml#L6

The tests were part of the pylint action (on push), your removed them from there. If you want to keep it that way, implement a new test action that runs on push.

Why the coveralls workflow does not run on push to the pull request branch, I don't know. Maybe "on" and "pull_request" need to be on the same line, as it was before

zain-sohail:
> > > 3. There is currently no pytesting done on push. This should be added, at least for one python version, for pushes to branches without a PR.
> > 
> > 
> > But this is the same as before. We were never using push. See
> > https://github.com/OpenCOMPES/sed/blob/19921fc33cfb4c491a490e547c61a100badd8128/.github/workflows/pytest_coveralls.yml#L6
> 
> The tests were part of the pylint action (on push), your removed them from there. If you want to keep it that way, implement a new test action that runs on push.


Oh yes I removed them. That's a fair point. I will put the single version test to be on every push then?
> Why the coveralls workflow does not run on push to the pull request branch, I don't know. Maybe "on" and "pull_request" need to be on the same line, as it was before

That shouldn't have mattered, from previous experience. I will check.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: "Setup Python, Poetry and Dependencies"
+        uses: packetcoders/action-setup-cache-python-poetry@main
+        with:
+          python-version: 3.8
+          poetry-version: 1.2.2
+
+      # Run pytest with coverage report and upload results to coveralls
+      - name: Run tests on python 3.8
+        run: |
+          poetry run pytest --cov=sed -n auto tests/
```
rettigl, line 26: This produces correct coverage for me locally: 
pytest --cov -n 10 -sv --show-capture=no tests/
Please add the other flags to track test output on the workflow, and remove the name from the cov flag to include tests into coverage calculation (at least that's our baseline)

zain-sohail: `poetry run pytest --cov=sed -n auto tests/`
gives
local:
Screenshot 2023-10-17 at 16 17 26

on github actions:
Screenshot 2023-10-17 at 16 20 17
also visible here: https://github.com/OpenCOMPES/sed/actions/runs/6533441246/job/17738634331


So I think that's also correct. But yes, adding the tests would be useful. --show-capture is useful to debug I'd say.
The problem seems not to be the coverage report but how the coveralls action is handling it. I am looking into that. It might be due to parallel pytest.

rettigl: Maybe just use the old way of uploading, and remove coverage from the multiversion test

rettigl: > --show-capture is useful to debug I'd say.

Yes, remove show capture, but maybe we should add --full-trace, this will not truncate trace backs on test failures, helpful to understand the cause

.github/workflows/testing.yml
```diff
@@ -0,0 +1,31 @@
+name: pytest
+
+# Triggers the workflow on push for all branches
+on: [push]
```
rettigl, line 4: The coveralls PR report does not get generated/updated if there is no testing work flow triggered on pull-request
That's why there was the testing-coveralls workflow before

zain-sohail:
Is it good now? @rettigl

rettigl:
> Is it good now? @rettigl
Three more things: 
- rebase
- add on: pull-request to the testing workflow
- remove line 653 from test_processor

rettigl:
Ah, and remove update_requirements

rettigl (approved):
LGTM now

zain-sohail:
> LGTM now

Thanks!

PR #192, opened 2023-10-18
Alterantive for dependabot

rettigl:
Dependabot is not the solution. It simply overwrites all limits to dependencies set in the pyproject.toml file. This is not what we want, just update the lock file.


zain-sohail:
> Dependabot is not the solution. It simply overwrites all limits to dependencies set in the pyproject.toml file. This is not what we want, just update the lock file.

I think a solution to this exists here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy with stragegy being lockfile-only

Secondly, using groups https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups would allow to have just one PR

rettigl:
Sounds promising, can you set this up and try it out?

PR #204, opened 2023-10-22
add wildcard pattern again

rettigl:


rettigl:
@zain-sohail this seems now to want to do the right thing, but always times out. Ideas?

zain-sohail:
Seems to me that in the monitored dependency files, there's also docs/requirements.txt. So that could be why it takes so long and times out. 
We can see if removing it helps. I think readthedocs can also work without the requirements.txt, from what I see online. Just probably needs to be configured properly.

PR #219, opened 2023-10-25
Dependency update action

rettigl:
Adds a github action to update the poetry.lock file weekly using a crone timed action. 
Sets up a pull request for review, if packages are updated, and lists the poetry update output in the PR description.
See previous test workflow runs and closed PRs for examples.


.github/workflows/update_dependencies.yml
```diff
@@ -0,0 +1,56 @@
...
+        uses: packetcoders/action-setup-cache-python-poetry@main
+        with:
+          python-version: 3.8
+          poetry-version: 1.2.2
+
+      # update poetry lockfile
+      - name: "Update poetry lock file"
+        id: update
+        run: |
+          poetry run poetry self update
```
zain-sohail, line 34: I dont think a poetry run is needed. poetry self update should work fine.

.github/workflows/update_dependencies.yml
```diff
@@ -0,0 +1,56 @@
...
+          exec 5>&1
+          UPDATE_OUTPUT=$(poetry run poetry update|tee >(cat - >&5))
```
zain-sohail, line 36: also here just poetry update would suffice

.github/workflows/update_dependencies.yml
```diff
@@ -0,0 +1,56 @@
...
+          echo "UPDATE_OUTPUT<<EOF" >> $GITHUB_OUTPUT
+          echo "$UPDATE_OUTPUT" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
+          STATUS=$(git status|tee >(cat - >&5))
+          echo "STATUS<<EOF" >> $GITHUB_OUTPUT
```
zain-sohail, line 42: I under that these statements are just capturing the packages that are updated.

.github/workflows/update_dependencies.yml
```diff
@@ -0,0 +1,56 @@
...
+          echo "$STATUS" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
```
zain-sohail, line 45: This command will allow us to have the requirements.txt updated

`poetry export --without-hashes --format=requirements.txt -o docs/requirements.txt -E docs -E notebook`

I believe you could put this as another step. and then we delete the generate-requirements.yml

zain-sohail (changes requested):
Is there a github actions run + PR that showcases the complete updating and merging behavior? 
I only found runs which tried updating but had nothing to update, and PRs which you closed without merging.

rettigl:
> Is there a github actions run + PR that showcases the complete updating and merging behavior? I only found runs which tried updating but had nothing to update, and PRs which you closed without merging.

https://github.com/OpenCOMPES/sed/actions/runs/6646883557/job/18061211259 was a successful run that produced this PR:
https://github.com/OpenCOMPES/sed/pull/216
I then updated the lockfile manually, and ran another run, 
https://github.com/OpenCOMPES/sed/actions/runs/6647035812/job/18061860224, which still changed some package names in the lock file. This created this PR: https://github.com/OpenCOMPES/sed/pull/218 which I merged.
Afterwards, I initiated this run: https://github.com/OpenCOMPES/sed/actions/runs/6647161677/job/18062064014 which now correctly does not issue a PR, because the lockfile was not updated.

.github/workflows/update_dependencies.yml
rettigl: We need to do this in the context of the poetry environment, no? Maybe its not needed, but it works

.github/workflows/update_dependencies.yml
rettigl: again, we need to act on the poetry environment. At least here I think we need the poetry run. Anyways, it works...

.github/workflows/update_dependencies.yml
rettigl: They are collecting the updated files in the repository, depending on which a PR is opened or not.

.github/workflows/update_dependencies.yml
rettigl: yes, this should work, I guess.

.github/workflows/update_dependencies.yml
rettigl: You are right, it's not needed

.github/workflows/update_dependencies.yml
rettigl: same here

.github/workflows/update_dependencies.yml
rettigl: done

zain-sohail (approved):
Looking quite good. I also like that it can be activated based on multiple criteria.

PR #222, opened 2023-10-26
fix small error and bypass push to main to test

zain-sohail:


zain-sohail:
Fixed this error but wondering if we can somehow only run the testing+coverage on only push or PR and not both. Is redundant if we do. 
TODO: remove on push(to all branches) for multitesting, and make it only for main

rettigl (approved):
Looks good, if you limit it to main again

rettigl:
We both missed that one it seems

rettigl:
> but wondering if we can somehow only run the testing+coverage on only push or PR and not both. Is redundant if we do.

I think this is necessary to get coverage summary in the PR. But you can investigate if you want. They run in parallel, so no real time issue.

rettigl:
We can use this PR also to update the badges in the Readme.md, something we also forgot

.github/workflows/testing_coverage.yml
```diff
@@ -1,7 +1,13 @@
...
-# Triggers the workflow on push and PR for all branches
-on: [push, pull_request]
+# Triggers the workflow on push for all branches and PR only for main
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches:
+      - main
```
rettigl, line 10: @zain-sohail: These changes you introduced after review here prevent the coverage report again from working for pull requests. This only works if the tests explicitly run for PRs in every branch. Please revert this.

rettigl: No, this means PRs _to_ main, no? Still, somehow many workflows don't run right now, and coverage reports don't work, I don't know why.

zain-sohail: Yes, that's indeed what it means. It should work on all pushes to whichever branch but since we always got it running twice when someone opened a PR before finishing the changes. 
This at least limits it to happen only on main. 
Can you give me examples where this workflow isn't running?

rettigl: https://github.com/OpenCOMPES/sed/pull/303 and https://github.com/OpenCOMPES/sed/pull/301 did not have coverage support. #303 Did indeed run the pull-request tests, but not the push tests, nor the linting.
In https://github.com/OpenCOMPES/sed/pull/295 it still worked, but also did not run push tests.
Not sure what the issue is.

rettigl: The reason is that coveralls does not find the base build. The reason behind this is that the respective branches are branched off the commits that bump the version after PR merges:
grafik
On these, you disabled running the tests, thus we don't have a base build in cases there are no other commits to main in-between. 
Solution: remove the exception of pyproject.toml from the test run trigger.

zain-sohail: I only turned it off since it was running twice but indeed, this seems like a valid reason to let the testing workflow run even with pyproject.toml change.
I will just push a change directly to main as this is an agreed upon change.

Issue #27, opened 2022-02-08
Include time-stamped external data


steinnymir:
Function to interpolate and map time-stamped external data into the data frame based on the time stamps of the data.

FHI already has some code, but needs to be wrapped into a function.

steinnymir:
@rettigl did we implement something in this direction?

rettigl:
Not yet, but We have code ready for this. I have this on my agenda for quite some time.

rettigl:
Addressed in #227

PR #227, opened 2023-11-01
Time stamped data

rettigl:
This PR adds code to add time stamped data to the dataframe using interpolation


rettigl:
Closes #27

steinnymir (approved):
Would be nice to see an actual test case. The tests are rather limited in showing use cases. Otherwise looks good, from what I can see without testing further.

rettigl:
> Would be nice to see an actual test case.

For the test data we have I unfortunately don't have meaningful timestamped data to add. I could maybe upload a second dataset.

PR #235, opened 2023-11-03
Build docs with github actions

zain-sohail:
Since these bokeh plots weren't working with our method, I am now using nbsphinx which directly executes the notebooks on it's own. But that won't be possible with readthedocs server due to timeout.

I looked at the docs building locally and it's really fast without tutorials so thought to implement the building directly and publish them to our github pages: 
https://opencompes.github.io/sed/

One benefit is we don't have to store the tutorial executed files on our repository like we are doing in #185

It takes a long time to build (50 mins) but it was taking about the same with our other method. Hence I propose we use this (first merging to the other branch and then main) 

Possibility to speed up: 

- Cache the WSe2 data (unlikely since the size is so large and it takes 6-10 mins currently)
- Cache additional dependencies for building
- Cache build files (that way it won't need to rebuild the whole docs everytime)


rettigl (approved):
I like this solution. Maybe adjust the triggers to work on push to the docs folder and to the notebooks only. Only think is that some of the figures are a bit wide (but scrollable).
We should also link the documentation on the readme.md, and update the docs badge and its link

Issue #211, opened 2023-10-25
Flash loader parquet file consistency check


rettigl:
Add a consistency check to the flash loader for the consistency of the intermediate parquet files with the config file settings, i.e. that all required columns are there. Additionally, add an option to force re-creation of the parquet files.

steinnymir:
yes, these features are really necessary. Thanks for making the issue! I continued forgetting...

Issue #212, opened 2023-10-25
Issue with SED installation on Maxwell via Conda approach


potorocd:
I installed SED on Maxwell to use the Tutorial notebook with test HEXTOF data using the 'Conda approach'. Simply following the installation doesn't lead to success. Here is a list of extra actions which were necessary for me:

1. To deal with virtual environments at the Maxwell cluster, I had to read this article ('conda activate' or 'conda init' doesn't work there):
https://rtd.xfel.eu/docs/data-analysis-user-documentation/en/latest/euxfel-software-environments/environments/
2. I could use env.yml once I removed 'locket==0.2.1' from the list of packages. Otherwise, it couldn't solve the environment.
3. List of packages which I had to install manually:
ipywidgets
fastdtw
lmfit
natsort
symmetrize
pynxtools
utils
4. To install a jupyter kernel ('python -m ipykernel install --user --name=sed_conda'), one also needs to install 'ipykernel'.
5. Eventually, I could load all modules with the required versions. However, there was an error: module 'pyarrow' has no attribute 'parquet'. I fixed it by using this hint:
https://github.com/modin-project/modin/issues/951

rettigl:
@steinnymir Why actually using conda for the non-poetry installation approach rather than pip? This is not so python standard. 
In my view, a more consistent setup can be achieved with virtualenv, and pip install -e ., which actually does not require an additional file, but directly uses the pyproject.toml (we removed the requirements.txt at some point for this reason, and that's what the current workflows use). The ipykernel still needs to be installed, though, I think.

steinnymir:
I totally agree. The conda approach is basically a residual from before poetry. `pip install -e .` would not have worked at the time, but should be the way to go now.

We should provide the commands to create an environment and then install with pip. Then we can still add ipykernel and the command to add the kernel to jupyter as an optional next step.

The conda approach, together with `env.yml` could be entirely removed.

We could also take the chance to move the setup instructions to the documentation page, where it should belong. We can make a link in the readme, and only leave a description of SED there.

zain-sohail:
We should definitely provide pip installation as an option. Pip also allows to install extras which exist as notebook extras in the toml file.
I think it's quite common practice to have the installation details in the README.md and also the documentation. So we could link the README in the docs.

By the way, I yesterday helped someone new to dima's group to install sed and she just followed the steps on the README for poetry and took a min to install, just copying and pasting commands.

rettigl:
This should now be obsolete, no?

PR #239, opened 2023-11-04
Flash buffer file consistency check and force creation

zain-sohail:
With reference to #211


sed/loader/flash/loader.py
```diff
@@ -18,7 +18,7 @@
 import dask.dataframe as dd
 import h5py
 import numpy as np
-import pyarrow as pa
+import pyarrow.parquet as pq
```
zain-sohail, line 21: this fixes an import error that was mentioned in #212

sed/loader/flash/loader.py
```diff
@@ -650,16 +650,32 @@ def buffer_file_handler(self, data_parquet_dir: Path, detector: str):
...
+        if not force_recreate:
+            # Check if the available channels match the schema of the existing parquet files
+            schemas = [pq.read_schema(file) for file in existing_parquet_filenames]
+            available_channels_set = set(self.available_channels)
+
+            for i, schema in enumerate(schemas):
+                schema_set = set(schema.names)
+                # Check if available_channels are the same as schema including pulseId
+                if schema_set != available_channels_set.union({"pulseId"}):
+                    raise ValueError(
```
steinnymir, line 692: This will always rise, as there are columns generated by the loader which are not in the config.

I tested with parquets generated before this PR and with ones generated in this PR. Both rise every time

steinnymir: I fixed this in a commit. I know its not good practice, but was easier to fix than to explain! :)

steinnymir (approved):
It seems to work now, after fixing the schema.
reloading works which is the most important feature, so LGTM!

steinnymir:
> With reference to https://github.com/OpenCOMPES/sed/issues/211

if you write: "closes #211" it will automatically close the issue on merge! ;)

sed/loader/flash/loader.py
```diff
@@ -178,6 +177,24 @@ def available_channels(self) -> List:
...
+    @property
+    def schema_from_config(self) -> set:
+        """Returns the set of all channels of the dataframe based on the config file
+
+        This is the complete list of channels which will be in the dataframe after creation.
+        """
+
+        schema = {"electronId", "trainId", "pulseId"}
+        if self._config["dataframe"]["split_sector_id_from_dld_time"]:
+            schema.add(self._config["dataframe"]["sector_id_column"])
```
rettigl, line 189: This can raise a key error if this entry is not defined. Use .get with default False.

sed/loader/flash/loader.py
```diff
@@ -178,6 +177,24 @@ def available_channels(self) -> List:
...
+        for channel in self.available_channels:
+            if channel == "dldAux":
+                for key in self._config["dataframe"]["channels"][channel]["dldAuxChannels"].keys():
+                    schema.add(key)
```
rettigl, line 193: Similar here. Aux channel does not have to be defined in the config.

sed/loader/flash/loader.py
```diff
@@ -625,17 +642,18 @@ def create_buffer_file(self, h5_path: Path, parquet_path: Path) -> Union[bool, E
             return exc
         return None
 
-    def buffer_file_handler(self, data_parquet_dir: Path, detector: str):
+    def buffer_file_handler(self, data_parquet_dir: Path, detector: str, force_recreate: bool):
```
rettigl, line 645: Type hints for returns missing (why?)

sed/loader/flash/loader.py
rettigl: It's maybe a matter of style, but why raise an error of they don't match? I would raise a warning, and recreate the files.

sed/loader/flash/loader.py
```diff
@@ -650,16 +668,45 @@ def buffer_file_handler(self, data_parquet_dir: Path, detector: str):
...
+                    missing_in_parquet = config_schema - schema_set
+                    if len(missing_in_parquet) > 0:
+                        missing_in_parquet_str = f"\nMissing in parquet: {missing_in_parquet}\n"
+                    else:
+                        missing_in_parquet_str = ""
+                    missing_in_config = schema_set - config_schema
+                    if len(missing_in_config) > 0:
+                        missing_in_config_str = f"\nMissing in config: {missing_in_config}\n"
+                    else:
+                        missing_in_config_str = ""
```
rettigl, line 695: Can you write a test for this?

rettigl (changes requested):
Generally looks good, but there are a few small things worth considering. Also, I would suggest to write a test for the new part.

sed/loader/flash/loader.py
steinnymir: True, and shoudl be for both "split_sector_id_from_dld_time" and "sector_id_column", even though redundant as they are interdependent.

sed/loader/flash/loader.py
steinnymir: At flash, the aux channel is always present, might consider adding it to the defaults

sed/loader/flash/loader.py
steinnymir: if the config is wrong, you don't want to overwrite the parquets. So in my opinion the error is necessary, not only a matter of style

sed/loader/flash/loader.py
```diff
@@ -178,6 +177,24 @@ def available_channels(self) -> List:
         available_channels.remove("pulseId")
         return available_channels
 
+    @property
+    def schema_from_config(self) -> set:
```
zain-sohail, line 181: Generally, the definition of this property seems redundant. We already have available_channels and get_channels_by_format. Adding the ids to them would lead to same result, except for sector_id. 
I am thinking to just extend the latter method to include all channels and change the name to get_channels

sed/loader/flash/loader.py
zain-sohail: perhaps we can just remove the split_sector_id_from_dld_time and use sector_id_column. If the column str exists, we can split the sector id. Otherwise we don't. So then there won't be a need for split_sector_id_from_dld_time bool

sed/loader/flash/loader.py
zain-sohail: Updated this

sed/loader/flash/loader.py
rettigl: I think this would be more difficult to understand the functionality then. It just requires a few more tests.

tests/loader/flash/test_flash_loader.py
```diff
@@ -0,0 +1,224 @@
...
+        fl.read_dataframe()
+
+
+def test_group_name_not_in_h5(config_file):
+    """
+    Test ValueError when the group_name for a channel does not exist in the H5 file.
+    """
+    config = config_file
+    config["dataframe"]["channels"]["dldPosX"]["group_name"] = "foo"
+    h5_path = "FLASH1_USER3_stream_2_run43878_file1_20230130T153807.1.h5"
```
rettigl, line 160: Maybe but this in the header

tests/loader/flash/test_flash_loader.py
```diff
@@ -0,0 +1,224 @@
...
+    - Check for an expected error related to the mismatch between available channels and schema.
+    - Force recreation of dataframe with the added channel, ensuring successful dataframe creation.
+    - Simulate a missing channel scenario by removing 'gmdTunnel2' from the configuration.
+    - Check for an error indicating a missing channel in the configuration.
+    - Clean up created buffer files after the test.
+    """
+    fl = FlashLoader(config=config_file)
+
+    # Read a dataframe for a specific run
+    fl.read_dataframe(runs=["43878"])
```
rettigl, line 187: This creates again a race condition with at least one other test in the reader test using the same generic paths in the config. Maybe we should overwrite them everywhere, if actual parquet files are being created

rettigl: It's in test_timed_dataframe

pyproject.toml
```diff
@@ -55,6 +55,7 @@ mypy = "^1.6.0"
 pycodestyle = "^2.11.1"
 types-pyyaml = "^6.0.12.12"
 types-requests = "^2.31.0.9"
+pyfakefs = "^5.3.0"
```
rettigl, line 58: What is this needed for?

rettigl (approved):
This looks good now, I have only a few very small comments left. Please merge after addressing these.

rettigl:
Ah, and add the `__init__.py` file to the flash folder.

pyproject.toml
zain-sohail: creating a fake file system that mimics maxwell. It's then used to test initialize paths

tests/loader/flash/test_flash_loader.py
zain-sohail: I will add the remove buffer at end of test_timed_dataframe
but I think we can think about introducing fixtures (or fixture factories I think) that allow for setting up test env and then a teardown (finish step)

PR #240, opened 2023-11-04
implement pump-probe calibration for hextof

steinnymir:
calibration methods for the pump-probe delay at FLASH.


steinnymir:
Main methods and tests present. Missing tutorial update

steinnymir:
ready for review!

steinnymir:
fixes and closes #246

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
         """Initialization of the DelayCalibrator class passes the config.
 
         Args:
             config (dict, optional): Config dictionary. Defaults to None.
         """
-        if config is None:
-            config = {}
-
-        self._config = config
+        self._config: dict = config or {}
```
rettigl, line 37: This kind of construct has raised an error in the tests, at least for list (with the sector delays). Not sure when and why this is a problem. I'd rather add one more line and make sure the code works.

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
-        self.adc_column = self._config["dataframe"]["adc_column"]
-        self.delay_column = self._config["dataframe"]["delay_column"]
-        self.calibration: Dict[Any, Any] = {}
+        self.adc_column: str = self._config["dataframe"].get("adc_column", None)
+        self.delay_stage_column: str = self._config["dataframe"].get("delay_stage_column", None)
+        if self.delay_stage_column is None and self.adc_column is None:
+            raise ValueError("No delay stage column specified.")
+        self.delay_column: str = self._config["dataframe"]["delay_column"]
+        self.calibration: Dict[str, Any] = self._config["delay"].get("calibration", {})
+        self.fluctuations: Dict[str, Any] = self._config["delay"].get("fluctuations", {})
```
rettigl, line 46: I'd suggest to make a similar contruct here as for the energy shifts, to create a shift dictionary, with constant and column-dependent shifts.

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
+        *args,
+        **kwargs,
+    ) -> Tuple[Union[pd.DataFrame, dask.dataframe.DataFrame], dict]:
+        """TODO: docstring"""
+        if self.loader not in ["mpes", "flash"]:
+            raise NotImplementedError(
+                f"Delay calibration is implemented for 'mpes' and 'flash', not {self.loader}.",
+            )
+        method = getattr(self, f"append_delay_axis_{self.loader}")
+        return method(df, *args, **kwargs)
```
rettigl, line 60: Is it necessary to do this overloading of the function? For flash, you do not strictly do a calibration (i.e. changing of coordinate system) of the delay axis, but essentially a shift, and this would be a function also interesting to have for mpes (if the calibration was off or so). So I would split the two into two seperate functions, calibrate_delay_axis, and shift_delay_axis.

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
+
+    def append_delay_axis_flash(
```
rettigl, line 62: shift_delay_axis

rettigl: and include the BAM here as well as column based shift

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
+            raise ValueError("No time0 value specified.")
+
+        if flip_time_axis is None:
+            fta_str: str = str(self.calibration.get("flip_time_axis", False))
+            flip_time_axis = fta_str.lower() in ["true", "1", "yes"]
+        else:
+            self.calibration["flip_time_axis"] = flip_time_axis
+
+        def calibrate_time(x, time0, flip_time_axis) -> Any:
+            return (x[delay_stage_column] - time0) * (-1 if flip_time_axis else 1)
```
rettigl, line 111: can you use the shift_by_constant function we already have in dfops? It can also do a flip, no?

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
+            and delay calibration metdata dictionary.
+        """
+        if len(kwargs) > 0:
+            print(f"WARNING: arguments {kwargs.keys()} are not used in flash delay calibration.")
+        assert self.loader == "flash", "Invalid loader for this method."
+        # pylint: disable=duplicate-code
+        if delay_stage_column is None:
+            delay_stage_column = self.delay_stage_column
+        if delay_column is None:
+            delay_column = self.delay_column
```
rettigl, line 95: These can now still be none, if they are not part of the default config.

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
+
+        if time0 is None:
+            time0 = self.calibration["time0"]
```
rettigl, line 98: can raise a key error

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
+        method = getattr(self, f"append_delay_axis_{self.loader}")
+        return method(df, *args, **kwargs)
+
+    def append_delay_axis_flash(
+        self,
+        df: Union[pd.DataFrame, dask.dataframe.DataFrame],
+        time0: float = None,
+        flip_time_axis: bool = None,
+        delay_stage_column: str = None,
+        delay_column: str = None,
```
rettigl, line 68: I would handle it the same as with the energy axis, that the default is to just shift in-place

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
+            meta=(delay_column, np.float64),
+        )
+
+        metadata: Dict[str, Any] = {
+            "applied": True,
+            "calibration": self.calibration,
+        }
+        return df, metadata
+
+    def append_delay_axis_mpes(
```
rettigl, line 126: append_delay_axis (or calibrate_delay_axis)

sed/calibrator/delay.py
```diff
@@ -174,6 +261,92 @@ def append_delay_axis(
 
         return df, metadata
 
+    def correct_delay_fluctuations(
```
rettigl, line 264: combine with shift_delay_axis, similar to shift_energy_axis

sed/config/default.yaml
```diff
@@ -95,6 +95,10 @@ delay:
   # value ranges of the analog-to-digital converter axes used for encoding the delay stage position
   # (in unbinned coordinates)
   adc_range: [1900, 25600]
+  # pump probe time overlap in ps
+  time0: 0
+  # if to flip the time axis
+  flip_time_axis: False
```
rettigl, line 102: add defaults for newly introduced dataframe columns, if you want to keep them

sed/calibrator/delay.py
```diff
@@ -25,22 +28,102 @@ class DelayCalibrator:
...
-
-        self._config = config
+        self._config: dict = config or {}
+        self.loader: str = self._config["core"]["loader"]
 
-        self.adc_column = self._config["dataframe"]["adc_column"]
-        self.delay_column = self._config["dataframe"]["delay_column"]
-        self.calibration: Dict[Any, Any] = {}
+        self.adc_column: str = self._config["dataframe"].get("adc_column", None)
+        self.delay_stage_column: str = self._config["dataframe"].get("delay_stage_column", None)
```
rettigl, line 41: I don't see any benefit of doing this. You don't to a check in the functions where you use these if the values are non-none, and the default-entries should always be there (unless a user purposefully overwrites the default config dict...)

sed/config/flash_example_config.yaml
```diff
@@ -55,6 +56,10 @@ dataframe:
   sector_id_column: dldSectorID
 
   sector_delays: [0., 0., 0., 0., 0., 0., 0., 0.]
+  # the corrected pump-probe time axis
+  delay_column: delay
+  # the delay stage column
+  delay_stage_column: delayStage
```
rettigl, line 62: Why do you want to separate them? Is this helpful to have both?

sed/core/processor.py
```diff
@@ -1268,6 +1268,8 @@ def align_dld_sectors(self, sector_delays: np.ndarray = None, **kwargs):
     # Delay calibration function
     def calibrate_delay_axis(
         self,
```
rettigl, line 1270: See above, split into two

sed/core/processor.py
```diff
@@ -1321,6 +1327,90 @@ def calibrate_delay_axis(
             else:
                 print(self._dataframe)
 
+    def correct_delay_fluctuations(
+        self,
```
rettigl, line 1331: combine with shift_delay_axis

sed/core/processor.py
```diff
@@ -1321,6 +1327,90 @@ def calibrate_delay_axis(
...
+            self._attributes.add(
+                metadata,
+                "correct_delay_fluctuations",
+                duplicate_policy="raise",
+            )
+
+    def save_delay_calibration(
+        self,
+        filename: str = None,
+        overwrite: bool = False,
```
rettigl, line 1382: That was something missing indeed, but there is also no loading for this for the mpes config yet. I'd suggest to change it similarly to the energy config as of now.

rettigl (commented):
Generally looks quite nice, but I find the overload mechanism overly complicated, and hard to understand and document. I'd suggest to split the code cleanly into a calibration part (for mpes), and a shift_delay_axis part, similar to the shift_energy_axis function we have, that can actually also be useful for mpes. That can also include the time jitter, so keeps the same number of processor functions.

sed/calibrator/delay.py
steinnymir: never gave me problems, not even in tests run locally. But I'll change it...

sed/calibrator/delay.py
steinnymir: removed

sed/calibrator/delay.py
steinnymir: agreed!

sed/config/default.yaml
rettigl: In the default config

steinnymir:
major refactor. Changed structure to mimic that of the enregy calibration.
Tests need to be updated.
Notebook works, only missing the right settings file which got lost...

steinnymir:
still missing tests for processor and flip from config does not work

sed/config/default.yaml
steinnymir: this is `sed/config/default.yaml`. isn't this the default config?

rettigl: Yes. Sorry, misinterpreted my own comment. What I mean is that you should add values for newly added dataframe columns to this file. E.g. delayStageAxis or how you called it. But if there is a mechanism that you don't need it, fine.

steinnymir: I think most of the features I implemented should correctly catch errors when relevant columns are missing. I prefer this in general, over a default setting which might be different from what expected.
I'd keep defaults for things like number of cores, histogramming options etc...

sed/core/dfops.py
```diff
@@ -281,7 +281,7 @@ def offset_by_other_columns(
         df (dask.dataframe.DataFrame): Dataframe to use. Currently supports only dask dataframes.
         target_column (str): Name of the column to apply the offset to.
         offset_columns (str): Name of the column(s) to use for the offset.
-        signs (int): Sign of the offset.
+        weights (int): weights to apply on each column before adding. Used also for changing sign.
```
rettigl, line 284: float

sed/core/dfops.py
```diff
@@ -304,13 +304,13 @@ def offset_by_other_columns(
     if any(c not in df.columns for c in offset_columns):
         raise KeyError(f"{offset_columns} not in dataframe!")
 
-    if isinstance(signs, int):
-        signs = [signs]
-    elif not isinstance(signs, Sequence):
-        raise TypeError(f"Invalid type for signs: {type(signs)}")
-    if len(signs) != len(offset_columns):
+    if isinstance(weights, int):
```
rettigl, line 307: float nor numeric or so

sed/core/processor.py
```diff
@@ -1280,7 +1279,7 @@ def add_energy_offset(
         Args:
             constant (float, optional): The constant to shift the energy axis by.
             columns (Union[str, Sequence[str]]): Name of the column(s) to apply the shift from.
-            signs (Union[int, Sequence[int]]): Sign of the shift to apply. (+1 or -1) A positive
+            weights (Union[int, Sequence[int]]): Sign of the shift to apply. (+1 or -1) A positive
                 sign shifts the energy axis to higher kinetic energies. Defaults to +1.
```
rettigl, line 1283: change docstring and type

sed/core/processor.py
```diff
@@ -1415,6 +1415,8 @@ def align_dld_sectors(self, sector_delays: np.ndarray = None, **kwargs):
     # Delay calibration function
     def calibrate_delay_axis(
         self,
+        time0: float = None,
+        flip_delay_axis: bool = False,
```
rettigl, line 1419: These can be removed now, no?

sed/core/processor.py
```diff
@@ -1424,6 +1426,8 @@ def calibrate_delay_axis(
         them from a file.
 
         Args:
+            time0 (float, optional): The pump-probe time overlap in picoseconds.
+            flip_delay_axis (bool, optional): Whether to flip the time axis direction.
```
rettigl, line 1430: remove

sed/core/processor.py
```diff
@@ -1460,6 +1464,8 @@ def calibrate_delay_axis(
 
                 self._dataframe, metadata = self.dc.append_delay_axis(
                     self._dataframe,
+                    time0=time0,
+                    flip_delay_axis=flip_delay_axis,
```
rettigl, line 1468: remove

sed/core/processor.py
```diff
@@ -1483,6 +1489,129 @@ def calibrate_delay_axis(
...
+    def add_delay_offset(
+        self,
+        constant: float = None,
+        flip_delay_axis: bool = None,
+        columns: Union[str, Sequence[str]] = None,
+        weights: Union[float, Sequence[float]] = None,
+        reductions: Union[str, Sequence[str]] = None,
+        preserve_mean: Union[bool, Sequence[bool]] = None,
+        inplace: bool = False,
+        rename: str = None,
```
rettigl, line 1524: I would suggest to rather than providing these, pass **kwds to allow such.

sed/core/processor.py
```diff
@@ -1483,6 +1489,129 @@ def calibrate_delay_axis(
...
+            weights (Union[float, Sequence[float]]): Weight of the shift to apply. A positive
+                sign shifts the delay axis to higher values. Defaults to +1.
+            preserve_mean (bool): Whether to subtract the mean of the column before applying the
+                shift. Defaults to False.
+            reductions (str): The reduction to apply to the column. Should be an available method
+                of dask.dataframe.Series. For example "mean". In this case the function is applied
+                to the column to generate a single value for the whole dataset. If None, the shift
+                is applied per-dataframe-row. Defaults to None. Currently only "mean" is supported.
+            inplace: Whether to apply the shift inplace or generate a new column.
+            rename: Whether to rename the column to "delay" after the shift.
```
rettigl, line 1540: The idea is that this is defined by the config normally.

rettigl: The energy offset also does not have this.

sed/core/processor.py
```diff
@@ -1483,6 +1489,129 @@ def calibrate_delay_axis(
...
+
+        Returns:
+            None
+        """
+        print("Adding delay offset to dataframe:")
+        delay_column = self._config["dataframe"]["delay_column"]
+        if delay_column not in self._dataframe.columns:
+            raise ValueError(
+                f"Energy column {delay_column} not found in dataframe! "
```
rettigl, line 1549: delay

sed/core/processor.py
```diff
@@ -1483,6 +1489,129 @@ def calibrate_delay_axis(
...
+                    rename=rename,
+                )
+            self._attributes.add(
+                metadata,
+                "add_delay_offset",
+                duplicate_policy="append",
+            )
+            self._dataframe = df
+            if self._timed_dataframe is not None:
+                self._timed_dataframe = tdf
```
rettigl, line 1587: This can be undefined, if the column does not exist! Rather, assign it directly as done in the other functions...

sed/core/processor.py
```diff
@@ -1501,7 +1630,8 @@ def save_workflow_params(
             self.save_energy_correction,
```
rettigl, line 1630: momentum correction is missing here (save_splinewarp)

rettigl (approved):
Thanks for implementing all the changes we discussed. There are a few places where I think some small inconsistencies still exist, which I marked. If you fix these, please go ahead and merge.

sed/calibrator/delay.py
```diff
@@ -174,6 +185,140 @@ def append_delay_axis(
...
+        df: dask.dataframe.DataFrame,
+        constant: float = None,
+        flip_delay_axis: bool = None,
+        columns: Union[str, Sequence[str]] = None,
+        weights: Union[float, Sequence[float]] = None,
+        preserve_mean: Union[bool, Sequence[bool]] = False,
+        reductions: Union[str, Sequence[str]] = None,
+        delay_column: str = None,
+        inplace: bool = True,
+        rename: str = None,
```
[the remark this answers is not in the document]
steinnymir: I am removing all references to inplace and rename. I might make a new PR adding this functionality in the same way everywhere.

sed/calibrator/delay.py
```diff
@@ -174,6 +185,140 @@ def append_delay_axis(
...
+                    "reduction": red,
+                }
+
+        # apply constant
+        if isinstance(constant, (int, float, np.integer, np.floating)):
+            df[delay_column] = df.map_partitions(
+                # flip sign if binding energy scale
+                lambda x: x[delay_column] + constant,
+                meta=(delay_column, np.float64),
+            )
```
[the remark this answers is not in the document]
steinnymir: as before, we do everything inplace for now.

sed/calibrator/energy.py
```diff
@@ -113,7 +113,7 @@ def __init__(
         self.color_clip = self._config["energy"]["color_clip"]
         self.sector_delays = self._config["dataframe"].get("sector_delays", None)
         self.sector_id_column = self._config["dataframe"].get("sector_id_column", None)
-        self.offset: Dict[str, Any] = self._config["energy"].get("offset", {})
+        self.offsets: Dict[str, Any] = self._config["energy"].get("offset", {})
```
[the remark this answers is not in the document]
steinnymir: renaming everything to offsets

sed/core/processor.py
steinnymir: This structure allows to use metadata duplicate_policy= raise correctly, without first overwriting the dataframe.

added a check to avoid the case when the column does not exist.

PR #248, opened 2023-11-07
Update build_deploy_docs.yml

steinnymir:
add a `verbose` option to the SedProcessor init, which toggles printing the dataframe after the calibration steps


rettigl (approved):
LGTM. You could add respective reporting code also in the functions that don't have it yet (e.g. energy shift, tof_ns axis), but also not so important.

PR #250, opened 2023-11-07
fix saving energy offset and add global save

steinnymir:
This PR adds the method to save energy offset parameters, and a method to call all save parameter methods in SedProcessor.

### DIFF
--- a/sed/calibrator/energy.py
+++ b/sed/calibrator/energy.py
@@ -9,6 +9,7 @@ from typing import Any
 from typing import cast
 from typing import Dict
 from typing import List
+from typing import Literal
 from typing import Sequence
 from typing import Tuple
 from typing import Union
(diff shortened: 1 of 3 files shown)


sed/calibrator/energy.py
```diff
@@ -1523,16 +1524,24 @@ def add_offsets(
                         signs.append(v["sign"])
                     except KeyError as exc:
                         raise KeyError(f"Missing sign for offset column {k} in config.") from exc
-                    preserve_mean.append(v.get("preserve_mean", False))
-                    reductions.append(v.get("reduction", None))
+                    pm = v.get("preserve_mean", False)
+                    if pm == "false":
+                        pm = False
+                    elif pm == "true":
+                        pm = True
```
rettigl, line 1531: Shouldn't the yaml library handle this correctly? Off course, these should not be strings in the yaml...

sed/calibrator/energy.py
```diff
@@ -1523,16 +1524,24 @@ def add_offsets(
...
+                    preserve_mean.append(pm)
+                    rd = v.get("reduction", None)
+                    if rd == "none":
```
rettigl, line 1534: maybe add .lower() ?

sed/calibrator/energy.py
rettigl: otherwhise, also add .lower() ?

rettigl (approved):
LGTM

sed/calibrator/energy.py
steinnymir: no, it does not automatically pick up bools and None

PR #258, opened 2023-11-12
ignore dependabot branches from triggering workflow

rettigl:
This should fix the problem in #256 by disabling the workflow for dependabot actions.


rettigl:
We could also limit this from only running on the main branch.

zain-sohail (changes requested):
I think switching to let this action run only on main is better.

zain-sohail (approved):

PR #264, opened 2023-11-13
SXP loader

rettigl:
PR for drafting an SXP loader


rettigl:
This should more or less work now, needs to be tested with a full run. For test data upload, I am waiting for David's approval.

rettigl:
This works for now. It has only a few modifications to the flash loader, but probably too many to become a single loader. 
The metadata stuff is untouched as of now, this needs to be updated once it's known how it works. Also, the additions from the flash consistency check can be ported once completed, but this I would do in separate PRs.

sed/loader/flash/utils.py
```diff
@@ -1,62 +0,0 @@
-"""Helper functions for the flash loader."""
```
zain-sohail, line 1: I think this is acciedently deleted

tests/data/loader/sxp/config.yaml
```diff
@@ -0,0 +1,97 @@
...
+    timeStamp: 's'
+    # energy: 'eV'
+    # E: 'eV'
+    kx: '1/A'
+    ky: '1/A'
+
+  channels:
+    timeStamp:
+      format: per_train
+      dataset_key: "/INDEX/timestamp"
```
zain-sohail, line 48: Not for this PR necessarily, but coming up with a schema for the dataframe in config file would be great. 
I think we had the discussion to use the group_names in mpes too.

tests/data/loader/sxp/RAW-R0016-DA03-S00000.h5
zain-sohail, line 1: Is this a test file?

tutorial/5_sxp_workflow.ipynb
zain-sohail, line 1: Lets wait for data approval before merging, I'd suggest

sed/loader/sxp/metadata.py
zain-sohail, line 1: It would be better to delete this in case of SXP and just leave the metadata method empty

sed/loader/sxp/loader.py
```diff
@@ -0,0 +1,891 @@
+# pylint: disable=duplicate-code
+"""
+This module implements the SXP data loader.
+This loader currently supports the SXP momentum microscope instrument.
+The raw hdf5 data is combined and saved into buffer files and loaded as a dask dataframe.
+The dataframe is a amalgamation of all h5 files for a combination of runs, where the NaNs are
+automatically forward filled across different files.
+This can then be saved as a parquet for out-of-sed processing and reread back to access other
+sed funtionality.
+Most of the structure is identical to the FLASH loader.
```
zain-sohail, line 10: I think just doing a replace all of flash to sxp would be good

sed/loader/sxp/loader.py
```diff
@@ -0,0 +1,891 @@
...
+from joblib import delayed
+from joblib import Parallel
+from natsort import natsorted
+from pandas import DataFrame
+from pandas import MultiIndex
+from pandas import Series
+
+from sed.core import dfops
+from sed.loader.base.loader import BaseLoader
+from sed.loader.sxp.metadata import MetadataRetriever
```
zain-sohail, line 33: can be removed

sed/loader/sxp/loader.py
```diff
@@ -0,0 +1,891 @@
...
+            paths and the parquet data directory path.
+
+        Raises:
+            ValueError: If required values are missing from the configuration.
+            FileNotFoundError: If the raw data directories are not found.
+        """
+        # Parses to locate the raw beamtime directory from config file
+        if (
+            "paths" in self._config["core"]
+            and self._config["core"]["paths"].get("data_raw_dir", "")
```
zain-sohail, line 73: I guess this would also be useful for flashloader

zain-sohail (changes requested):
Generally I have seen the git blame so I got the idea where you made changes.
Is this going to be a permenant loader or it's just for your beamtime?
Because adapting flash loaders structure makes this confusing and likely unmaintainable. We should later definitely think of a solution to this.

sed/loader/flash/utils.py
rettigl: No, I moved the content (one function) into the general utils, as it is also used in the sxp loader now.

tests/data/loader/sxp/config.yaml
rettigl: This kind of defines a schema, but yes, reworking the mpes structure here is on my todo (not very high up though).

tests/data/loader/sxp/RAW-R0016-DA03-S00000.h5
rettigl: Its a few trains from one of the runs. I clarified this with Markus Scholz.

tutorial/5_sxp_workflow.ipynb
rettigl: Not sue there will be a public example dataset soon, but I'd still like to merge this and make it usable easily.

sed/loader/sxp/metadata.py
rettigl: yes, I can install a dummy.

sed/loader/sxp/loader.py
rettigl: I put this there on purpose ;) ...

rettigl: this is the only ocurrence, btw

sed/loader/sxp/loader.py
rettigl: yes

sed/loader/sxp/loader.py
rettigl: There are a few things that seem inconsistent to me still. E.g. I think it would be good if you could auto-find the beamtime-dirs, and still user-define a parquet-dir. But that's something for another PR, I think.

rettigl:
> Generally I have seen the git blame so I got the idea where you made changes. Is this going to be a permenant loader or it's just for your beamtime? Because adapting flash loaders structure makes this confusing and likely unmaintainable. We should later definitely think of a solution to this.

Well, In gerneral the requirements for FLASH and SXP are very similar, i.e. the train /micro/macrobunch structure etc., so it does not make sense to me to deviate too much from what we have. But I agree we need to find a way to maintain this. Either we outsource the common code into utils files, or build an intermedate FEL loader, where FLASH and SXP loader inherit from.

sed/loader/sxp/loader.py
zain-sohail: I put this in #281

zain-sohail:
> > Generally I have seen the git blame so I got the idea where you made changes. Is this going to be a permenant loader or it's just for your beamtime? Because adapting flash loaders structure makes this confusing and likely unmaintainable. We should later definitely think of a solution to this.
> 
> Well, In gerneral the requirements for FLASH and SXP are very similar, i.e. the train /micro/macrobunch structure etc., so it does not make sense to me to deviate too much from what we have. But I agree we need to find a way to maintain this. Either we outsource the common code into utils files, or build an intermedate FEL loader, where FLASH and SXP loader inherit from.

I think best solution is to define different classes for these different aspects and create the actual loaders to instantiate these classes

zain-sohail (approved):

PR #274, opened 2023-11-16
Rework hextof tutorial notebook

### LINKED ISSUE #251: Externally accessing flash public data (for docs)
As the notebook is restricted to maxwell, we can not do the execution on github actions and put those changes in the documentation. Of course, we can have the notebook unexecuted but thought other solutions should be discussed.

Executing and putting it on the repo isn't a good idea as notebooks generally don't well with version control.

steinnymir:
Added explaination of all steps, and full workflow currently available for HEXTOF@FLASH.

This is a simple PR, which changes nothing in fucntionality. I'd love to merge this quickly.

Closes #251


rettigl:
It would make sense to solve the data download questions here as well, but if time is an issue, we can also postpone this.

steinnymir:
Upgrading this for use in Maxwell is now more urgent than the data problem. I'd like to merge this as is

tutorial/4_hextof_workflow.ipynb
```diff
@@ -46,14 +53,33 @@
    "source": [
     "local_path = Path(sed.__file__).parent.parent/'tutorial/'\n",
     "config_file = local_path/'hextof_config.yaml'\n",
-    "assert config_file.exists()"
+    "assert config_file.exists()\n",
+    "config_override = {\n",
+    "    \"core\": {\n",
+    "        \"paths\": {\n",
+    "            \"data_raw_dir\": \"/asap3/flash/gpfs/pg2/2023/data/11019101/raw/hdf/offline/fl1user3\", \n",
+    "            \"data_parquet_dir\": \"/home/agustsss/temp/sed_parquet/\",\n",
```
rettigl, line 61: This should not point to your home folder. Maybe let's use some relative path, assuming the sed package is installed in the home folder, and this file resides in the tutorials

tutorial/4_hextof_workflow.ipynb
```diff
@@ -62,11 +88,199 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# the name of the axes on which we want to bin\n",
+    "axes: List[str] = ['dldPosY', 'dldPosX']\n",
+    "# the number of bins for each axis\n",
+    "bins: List[int] = [480, 480]\n",
+    "# for each axis, the range of values to consider\n",
+    "ranges: List[List[int]] = [[420,900], [420,900]]\n",
```
rettigl, line 167: Strictly speaking, they should be a list of tuples of float, but this works off course.

tutorial/4_hextof_workflow.ipynb
```diff
@@ -62,11 +88,199 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sp_t05 = SedProcessor(runs=[44798], config=config_override, user_config=config_file, collect_metadata=False)\n",
+    "sp_t05.add_jitter()\n",
+    "sp_t10 = SedProcessor(runs=[44799], config=config_override, user_config=config_file, collect_metadata=False)\n",
+    "sp_t10.add_jitter()"
```
rettigl, line 216: You could use the same processor reference sp if you do the binning directly.

tutorial/4_hextof_workflow.ipynb
```diff
@@ -108,27 +322,28 @@
...
   {
-   "cell_type": "code",
-   "execution_count": null,
+   "cell_type": "markdown",
    "metadata": {},
-   "outputs": [],
    "source": [
-    "plt.figure()\n",
-    "res.plot.line(x='dldTime')"
+    "We binned not only in `dldTime` but also in `sampleBias`. This allows us to separate the spectra obtained at different bias values. "
```
rettigl, line 328: I suggest to add a cell or two explaining the overview histogram function here (view_event_histogram). With that you can easily inspect the values of sampleBias sampled, and the dldTime range of interest.

tutorial/4_hextof_workflow.ipynb
```diff
@@ -347,6 +600,13 @@
     "sp.add_energy_offset()"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## plot the delayStage values"
```
rettigl, line 607: This does only output noise for me. Probably something goes wrong with energy calibration/offsetting, but I don't have a clue. 
In any case, if you run this from the tutorials folder, it will pick up parameters from the sed_config.yaml file, which is there by default, so either we remove this, or disable the folder_config in any processor call here. This is, however, not the reason for this energy calibration problem, becaue it persists if I remove the file.

tutorial/4_hextof_workflow.ipynb
```diff
@@ -525,6 +812,13 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "if we look at the fully corrected dataframe, there are 76 graph layers. This is too much, and is what makes the binning procedure slow. Optimization of this is in the future plans."
+   ]
   }
  ],
  "metadata": {
```
rettigl, line 824: Add a saving cell (to h5 or tiff).

rettigl (approved):
I like the basic instructive rewrite, so general aproval. Some things, however, do not work for me, or could be improved, see comments.

tutorial/4_hextof_workflow.ipynb
steinnymir: I removed the sed_config.yaml, as it does not belong there. Also this has never been a problem for me as the work directory is my home, not that folder. 
I cannot reproduce the problem you mention, my plotting works just fine. It would be good if you could identify the source of the problem, as it is likely in the functionalities of the calibration/corrections.

steinnymir: There is also an other problem which I still have not figured out, and that is why the energy calibration adds the correct offset when calculated, and when applied after saving the parameters, all is offset by 2eV

rettigl: Seems to work now, cannot reproduce the problem. The issue with the 2 eV I think come from the the fact that in the one case you offset for the mean of your bias serias, which spans about 4 eV, so it is about 2 eV smaller than the higher bias, which I think was used for the latter experiment. So for that case, you experience that 2 eV offst. 
This is what I want to get rid off by taking the actial bias voltage into account in the energy calibration, and subtracting it there already. Then you can add the full value of the bias, and don't have to subtract the mean in the shift function, which produces the problem in the first place.

rettigl:
Another thing to add would be the histogram normalization, maybe.

tutorial/4_hextof_workflow.ipynb
steinnymir: > Seems to work now, cannot reproduce the problem. The issue with the 2 eV I think come from the the fact that in the one case you offset for the mean of your bias serias, which spans about 4 eV, so it is about 2 eV smaller than the higher bias, which I think was used for the latter experiment. So for that case, you experience that 2 eV offst.

Does this mean that setting ref_id to reference the bias which will be used later will set the correct energy calibration parameters for that ref_id?

> This is what I want to get rid off by taking the actial bias voltage into account in the energy calibration 

This would be great. We could even add an arbitrary number of columns, which then, when adding them as offset, we don't need to compensate the mean.

tutorial/4_hextof_workflow.ipynb
steinnymir: I agree and will add this. Also, I never tried it and I better know how it works before the BT

steinnymir:
> Another thing to add would be the histogram normalization, maybe.

Absolutely!

tutorial/4_hextof_workflow.ipynb
```diff
@@ -216,18 +503,17 @@
...
   {
-   "cell_type": "code",
-   "execution_count": null,
+   "cell_type": "markdown",
    "metadata": {},
-   "outputs": [],
    "source": [
-    "sp.dataframe"
+    "## correct offsets\n",
+    "The energy axis is now correct, but still the curves do not stack on eachother as we are not compensating for the `sampleBias`. In the same way, we can compensate the photon energy (`monocrhomatorPhotonEnergy`) and the `tofVoltage` "
```
zain-sohail, line 516: running this cell, I get the error:
`KeyError: "['sampleBias', 'monochromatorPhotonEnergy', 'tofVoltage'] not in dataframe!"`

tutorial/sed_config.yaml
zain-sohail, line 1: there's a file called hextof_config.yaml in this folder

tutorial/4_hextof_workflow.ipynb
zain-sohail: It seems you updated the flash_example_config whereas I thought the tutorial was using hextof_tutorial.yml

sed/config/flash_example_config.yaml
```diff
@@ -3,20 +3,26 @@
 core:
   # defines the loader
   loader: flash
-
   # the beamline where experiment took place
   beamline: pg2
-
-  # The paths to the raw and parquet data directories.
+  # the instrument used
+  instrument: hextof
```
zain-sohail, line 9: We dont really use the instrument parameter anymore

sed/config/flash_example_config.yaml
```diff
@@ -3,20 +3,26 @@
...
+    # location of the intermediate parquet files.
     data_parquet_dir: "tests/data/loader/flash/parquet"
 
-  # These can be replaced by beamtime_id and year to automatically
-  # find the folders on the desy cluster
-
-  # beamtime_id: xxxxxxxx
-  # year: 20xx
+binning:
+  # Since this will run on maxwell most probably, we have a lot of cores at our disposal
```
zain-sohail, line 24: I believe including the config file link from documentation in the notebook would be good. As this file isnt really accesible to users so trivially (when they install with pip)

sed/config/flash_example_config.yaml
```diff
@@ -139,6 +182,8 @@ dataframe:
   # (Not to be changed by user)
   beamtime_dir:
     pg2: "/asap3/flash/gpfs/pg2/"
+    hextof: "/asap3/fs-flash-o/gpfs/hextof/"
```
zain-sohail, line 185: again, adding this here right now doesnt make sense I think

zain-sohail (changes requested):
Very thoroughly explained. I don't think a new user would have any issues getting into this workflow.

tutorial/4_hextof_workflow.ipynb
zain-sohail: I faced this problem since I am using the sed-processor 0.1.5 which doesnt yet include this change.

tutorial/sed_config.yaml
steinnymir: there should no longer be one after this PR

zain-sohail: Its stll in this branch in tutorial folder

tutorial/4_hextof_workflow.ipynb
zain-sohail: Yes right.

tutorial/sed_config.yaml
rettigl: you are talking about different things. the sed_config will be removed, hextof_config stays (and is used for this tutorial, see cell 2)

sed/config/flash_example_config.yaml
rettigl: The flash example config should be also included if you install via pip

zain-sohail:
Deployment stopped because only main branch is allowed to deploy but I overrid it for checking. I think this is good to merge.

rettigl:
Yes, it's clear that the deploy did not work. I downloaded the generated artefact and checked locally that it was fine.
However, before merging there are still several issues and todos to fix in the documentation.
@steinnymir @zain-sohail please feel free to finalize

tutorial/4_hextof_workflow.ipynb
rettigl: I added respective cells

rettigl:
> > Another thing to add would be the histogram normalization, maybe.
> 
> Absolutely!

added

sed/config/flash_example_config.yaml
rettigl: removed

sed/config/flash_example_config.yaml
rettigl: @zain-sohail can you please set this here and in the notebook to something useful and resolve? I don't have any idea about the fs structure on maxwell, nor where I could have write access.

zain-sohail (approved):

PR #276, opened 2023-11-16
Add division model k-calibration

steinnymir:
Add the simpler momentum calibration method proposed by @balerion.


steinnymir:
Change naming to reflect the actual behaviour.

This is a distortion correction, not  a k calibration method

zain-sohail:
@balerion would you like to generate the parameters for it? This PR is stuck otherwise

Issue #279, opened 2023-11-17
add path-ignore to workflows


zain-sohail:
running the release when there are no changes to the package doesn't make sense
likewise running the documentation when tutorials, sed etc haven't changed also is pointless

So issue to restrict these

PR #280, opened 2023-11-17
Automation improvements

zain-sohail:


zain-sohail:
closes #279

rettigl:
The workflows should also run if one updates the workflow files...

zain-sohail:
> The workflows should also run if one updates the workflow files...

Shouldn't be the case. The workflows don't check or do anything with the workflow files. Most of our workflows are based on checking sed or tests folder (so they can be further restricted). And documentation is based on sed and tutorial. Build is based on toml and sed.

If you mean we should be able to test the workflows, I'd suggest we add a manual run option.

rettigl:
What I mean is that a workflow should run if itself is being updated. Today, the documentation workflow was broken, and it did not run when I fixed it.

zain-sohail:
I get that. I can add for each workflow this option. 

Currently pytest and pylint run on all pushes. Shall that also be restricted to certain folders?

PR #291, opened 2023-11-21
Fix to release workflow

zain-sohail:
In reference to #285

Some comments in #280 are also addressed here

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,7 @@ name: Publish to PyPI
 # and a release job on every tag push.
 # A publish job is executed on every successful prerelease or release job
 # And if publish is successful, the version is also updated in the pyproject.toml file and pushed to main branch
-  # This workflow only changes the pyproject.toml which is in ignore path, so workflow does not trigger itself
+  # Workflow does not trigger itself as it only changes pyproject.toml, which is not in paths for this workflow
 # The package is distributed as sed-processor
 on:
   push:
@@ -15,8 +15,6 @@ on:
     paths:
       - sed/**/*
       - .github/workflows/release.yml
-    paths-ignore:
-      - pyproject.toml
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,8 +4,7 @@ name: Publish to PyPI
 # and a release job on every tag push.
 # A publish job is executed on every successful prerelease or release job
 # And if publish is successful, the version is also updated in the pyproject.toml file and pushed to main branch
-  # By default, a workflow run that is triggered by a push event will
-  # not trigger another push event when it pushes changes to the repository.
+  # This workflow only changes the pyproject.toml which is in ignore path, so workflow does not trigger itself
 # The package is distributed as sed-processor
 on:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -15,7 +14,12 @@ on:
       - '*'
     paths:
       - sed/**/*
+      - .github/workflows/release.yml
+    paths-ignore:
       - pyproject.toml
```
rettigl, line 19: https://github.com/OpenCOMPES/sed/actions/runs/6942391340
says this is not valid syntax

zain-sohail: Right. I just put it for extra certainty but eitherways the workflow won't trigger since it only ever changes the toml file and that's not in the paths.

.github/workflows/release.yml
```diff
@@ -132,6 +134,13 @@ jobs:
     if: always() && (needs.publish.result == 'success')
     runs-on: ubuntu-latest
     steps:
+      - name: Generate a token
+        id: generate_token
+        uses: actions/create-github-app-token@v1
+        with:
```
rettigl, line 140: How is this different from tibdex/github-app-token@v1, which I used for the update_dependencies workflow?

.github/workflows/release.yml
```diff
@@ -143,15 +152,13 @@ jobs:
           name: pyproject
           path: sed-processor
 
-      # new_branch set for cases when tagging a release
       - name: Commit changes
+        env:
+          GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
```
rettigl, line 157: Which permissions will this token have? Does it allow to bypass the pull request rule?

rettigl (approved):
We can try it out, not sure I understand if it should work or not.

rettigl:
Did not work, was it because the version was out of date with pipy?
I'd suggest to make a new main release 1.6, how do I do that?

.github/workflows/release.yml
zain-sohail: This is the official one provided by github. The other one was used before they had official one

.github/workflows/release.yml
zain-sohail: The token has the ones you set before in the app. The only thing that's different is that I allow the app to bypass branch protection. You can see that if you check the branch protection rules

zain-sohail:
> Did not work, was it because the version was out of date with pipy? I'd suggest to make a new main release 1.6, how do I do that?

It's because last time the publish job worked but the bump version job didn't. So next time the bump job would be a version behind. Easy fix is to just bump the version once manually.

git tag v0.1.6 and pushing the tag will trigger the workflow to update to version 0.1.6

rettigl:
Still not working

zain-sohail:
> Still not working
Fixed now. 
But I think we should include paths-ignore pyproject.toml to all workflows since otherwise they will execute twice for no reason. Once for actual PR and once after bumping version.

rettigl:
Thanks, great. Also great development guide. Do we have some mechanism to restrict anyone from creating new releases?
When I pushed my created tag, it pushed all kind of tags with version numbers >0.2, no idea where they came from. Can you clean that up?

zain-sohail:
> Thanks, great. Also great development guide. 

Thanks. Also thinking to extend for general contributions but also great if we extend it for developing loaders, as an API reference is clearly not enough.

> Do we have some mechanism to restrict anyone from creating new releases? 

Ok just added v* to protected tags. But currently, it's just a very few of us with write access to the repo. But still a good idea.

> Do we have some mechanism to restrict anyone from creating new releases? When I pushed my created tag, it pushed all kind of tags with version numbers >0.2, no idea where they came from. Can you clean that up?

The other versions were my tests on testpypi. You can see those are from a few days ago. I will remove them

Issue #142, opened 2023-08-29
df_partitions as list of df partitions


rettigl:
Allow df_partitions kw argument of the compute and pre_binning methods of the processor to accept also lists/slices of dataframe partition ids, instead of just a number with the number of df partitions to use

rettigl:
This has been working all-along. PR #298 adds some unification to the type hints.

PR #298, opened 2023-11-21
unified typing of df_partitions

rettigl:
This implements a small unification of the code addressing #142


zain-sohail (approved):
Looks good. I guess that instead of casting, it'd be better to make the int into a list or another sequence.

rettigl:
> Looks good. I guess that instead of casting, it'd be better to make the int into a list or another sequence.

Had the same idea yesterday evening, will do so.

PR #305, opened 2023-11-29
BTHofmann2023 additions

steinnymir:
Bug fixes and utilities added during the beamtime at FLASH in nov2023.

### DIFF
--- a/sed/binning/binning.py
+++ b/sed/binning/binning.py
@@ -293,7 +293,8 @@ def bin_dataframe(
         xarray object, combining the data with the axes (bin centers).
     """
     bins, axes, ranges = simplify_binning_arguments(bins, axes, ranges)
-
+    # filter dataframe to use only the columns needed for the binning
+    df = df[axes]
     # create the coordinate axes for the xarray output
     # if provided as array, they are interpreted as bin centers
     if isinstance(bins[0], np.ndarray):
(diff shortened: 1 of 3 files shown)


rettigl:
> * filter the dataframe before binning to only look at the required columns. This avoids errors when there are issues with values in other columns (i.e. if a column is entirely nans)

I don't quite understand this issue, and as I have commented, it makes computation of the dataframe, and thus binning, slower. Can you elaborate on this problem? I tried to reproduce by adding a NaN column to the df, which worked without problems...

steinnymir:
> it makes computation of the dataframe, and thus binning, slower.

did you test this? does it really make it slower? I cannot recall exactly what the problem was, but will get back to it during analysis certainly. The problem was with a column which had dtypes not accepted by numba. I don't know why numba needed to test all dtypes of all columns, but this was a quick fix which solved the problem.

rettigl:
> > it makes computation of the dataframe, and thus binning, slower.
> 
> did you test this? does it really make it slower?
> 
> I cannot recall exactly what the problem was, but will get back to it during analysis certainly. The problem was with a column which had dtypes not accepted by numba. I don't know why numba needed to test all dtypes of all columns, but this was a quick fix which solved the problem.

For mpes, I get something like this pretty consistently:
grafik
And also binning is notably slower. For Flash dataframes, the difference is smaller, but also there.
grafik
Note the added graph layer:
grafik

rettigl:
Tests for the __repr__ methods are failing. This is due to various reasons, the first being that you don't check if the dataframe is loaded, another that you don't check if the columns you request exist (e.g. for non-Flash-loaders).

PR #314, opened 2023-12-19
Sxp beamtime additions

rettigl:
Fixes and modifications to the SXP loader implemented during the beamtime in November
Tutorial data are not publically available for now


tutorial/sed_config.yaml
```diff
@@ -1,9 +1,9 @@
 energy:
   calibration:
-    E0: -31.09210508670466
-    d: 1.0954452312102088
+    E0: -196.10237272644352
```
zain-sohail, line 3: I guess these changed as the sed_config is overwritten by the save parameters method

zain-sohail (approved):
I can't really test this update without any test data but if this works and is not failing any general tests, we can merge this. 
Later we can combine the functionality with #311

tutorial/sed_config.yaml
rettigl: This file should have been deleted already anyways. I will remove it here now.

PR #330, opened 2024-01-18
complete config model for flash/SXP loaders

zain-sohail:


rettigl:
I don't understand this PR. It appears to just add the config model, without it being used somewhere in the code. Where are these changes?

Issue #247, opened 2023-11-07
Upgrade python version


steinnymir:
The latest release of python is now 3.12, which means our 3.8 is now 4 versions back, and we should consider upgrading.

I would not go all the way to 3.12 which was just released, but I'd rather push to 3.10.

according to https://devguide.python.org/versions/ 3.10 is now not recieving any updates rather than security, while 3.11 and 3.12 are still recieving bugfixes.

I tested 3.10 and at least for the hextof tutorial notebook and the tests, it worked just fine.

Also, I'm a big fan of 3.10, as it greatly improves readability of type hihts and their flexibility.
No need for Union or capitalized List Dict etc...
```python
def foo(x: int | list[int]) -> None:
   if not isinstance(x, int | float):
       raise ValueError()
```
There are also other performance improvements and a case funtion which is nice
https://docs.python.org/3/whatsnew/3.10.html

On the other hand:
> Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite. See [Faster CPython](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-faster-cpython) for details.

rettigl:
I am absolute up for making sure our code is compatible with recent python version, so we should extend our multiversion tests accordingly. However, I would keep compatibility to python 3.8, because it is the default python shipped with many long-term support os-es, e.g. ubuntu 20.04, which has support until 2025.

steinnymir:
I wonder if using the new features of python 3.10 would break compatibility with 3.8.
I have the feeling I did this and it worked, but i need to test to be sure

zain-sohail:
Using the 3.10+ specific features definitely for example cause linting errors with other versions.
If we don't change the code, there shouldn't be any problem with moving to 3.10, as all our tests already show it works fine. That means that the dependencies we have are all compatible with 3.10.

zain-sohail:
Just my comments. I am totally okay with moving to 3.10. We can change our linting workflow to adapt to it as well.

steinnymir:
I tested functions using 3.10 functionalities in 3.8, and it obviously does not work... 😭 

We probably need to stick to developing in the oldest version we want to keep compatibility with, which is a pitty.
I would therefore propose we find really good reasons why to stick with 3.8 and not upgrade... the "its the default in an os" does not feel like a good enough reason to me.

zain-sohail:
[...]

PR #337, opened 2024-01-30
Add support for python 3.11

rettigl:
This PR adds support for python 3.11
To achieve this, several package restrictions have been released.
Python 3.12 is still not supported by (at least) numba.


.github/workflows/testing_multiversion.yml
```diff
@@ -1,17 +1,15 @@
-name: unit tests [Python 3.8|3.9|3.10]
+name: unit tests [Python 3.8|3.9|3.10|3.11]
 
 on:
   push:
-    branches: [ main ]
-    paths-ignore:
```
zain-sohail, line 6: Any reason this path was removed? It would end up causing a recursive action run.

pyproject.toml
```diff
@@ -13,61 +13,61 @@ keywords = ["sed", "mpes", "flash", "arpes"]
...
+fastparquet = ">=0.8.0"
+h5py = ">=3.6.0"
 ipympl = ">=0.9.1"
 ipywidgets = ">=7.7.1"
-lmfit = "^1.0.3"
-matplotlib = "^3.5.1"
-natsort = "^8.1.0"
-numba = "^0.55.1"
-numpy = ">=1.18,<1.23"
+lmfit = ">=1.0.3"
```
zain-sohail, line 24: Is there a reason to changing `^` to `>=` ?

Poetry specification recommends https://python-poetry.org/docs/dependency-specification/ and also when we do poetry add <package-name>, it automatically uses `^`

.github/workflows/testing_multiversion.yml
```diff
@@ -1,17 +1,15 @@
-name: unit tests [Python 3.8|3.9|3.10]
+name: unit tests [Python 3.8|3.9|3.10|3.11]
 
 on:
   push:
-    branches: [ main ]
```
zain-sohail, line 5: If we are running this on every branch push, probably running the coveralls can also be done here.

rettigl: I think I just removed this for testing purposes, will put it back, thanks

.github/workflows/testing_multiversion.yml
rettigl: see above

pyproject.toml
rettigl: Using ^ will prevent any update of the highest version in the hierarchy, e.g. and update from 0.x.y to 0.x+1.y will be prevented. While this might be a good idea for packages following the major.minor.patch versioning style, many packages have adopted a year.month.day style of versioning, where ^essentially suppresses most updates, at least ones in a new year.
As many packages needed update for python 3.11, I essentially opted to remove any upper limits. I think we have good enough tests to recognize any incompatibilities, allowing us to add respective limits in time.

zain-sohail:
closes #247

PR #341, opened 2024-02-05
Add ascale parameter

### LINKED ISSUE #144: Add ascale to momentum corrector interface and config for anisotropic systems
Add the mc.ascale parameter as parameter to the interface, and saving options with respective config dict entries, together with documentation, examples and tests. This is required for anisotropic (e.g. orthorhombic) systems.

### LINKED ISSUE #343: Documentation for sp.calibrate_energy_axis
Fitting the energy calibration with the law d^2/(t-t0)^2 is the most appropriate way and leads to the best results.

However, the function is strongly peaked around t0 and can be unstable against initialization parameters (fit might not converge at all).

In sed.calibrate.energy "fit_energy_calibation" this control is actually implemented in a nice way, with a dictionary input to control not just initial value but also min, max and "lock" (whether the parameter is varied or not during the optimization).

It would be of great benefit to the user, and I think a relatively frequent use case, if this was explained in the documentation of sp.calibrate_energy_axis.

rettigl:
Adds an interface to the ascale parameter. Still needs:
- [x] Tests
- [x] A tutorial for documentation

Closes #144 

It also contains a description for the parameter fixing of the energy calibration, and closes #343


zain-sohail (approved):
I have tested locally and the notebook decribes the process very well. Code looks good to merge.

Issue #345, opened 2024-02-20
Problem with get metadata from SciCat


kutnyakhov:
Get metadata from scicat using ingestor login and the password is not functioning after scicat was migrated to the new backend.
With the help of IT guys, I could adopt /sed/sed/loader/flash/metadata.py file to use my own token generated by SciCat.
Also config.yaml file has to be changed with metadata:scicat_token:"eyJhbGciOi...." #copied from the SciCat front-end webpage (User/Settings/SciCat Token).
Thus: 
- pros - no need to make public credentials of ingestor;
- cons - token is not a long leaving thing (probably just a few hours) and has to be recopied from SciCat
Somehow I didn't manage to push changes to github directly, so metadata file is attached.
[metadata.txt](https://github.com/OpenCOMPES/sed/files/14345312/metadata.txt)

rettigl:
This is probably something to discuss with the admins of that server. How is it handled for automatic igestion, e.g. from the DAQ? What account is being used there? A r/o accound would be sufficient for our use case. If there is no permanent solution, then maybe a cell that sets the respective config item in a template notebook would be a good idea, where users can copy paste their Token.

kutnyakhov:
I was trying to make the local branch get-metadata-from-scicat-using-token and push it on github, but got 403 error :)

Concerning lifetime - let's see what will be an answer from scicat guys, but I could not find any easy set for this via the web interface.

zain-sohail:
> I was trying to make the local branch get-metadata-from-scicat-using-token and push it on github, but got 403 error :)

It seems you weren't having write permissions. I have added you now. But otherwise, I also made a PR #347 to address what you put

zain-sohail:
Now the token can be added as a keyword argument when calling the sed processor.

kutnyakhov:
After discussing with ITs, it seems that we can also use the build conda package for SciCat
conda install -c https://hasfcpkg.desy.de/doocsconda scicat
which already exists as the default package inside of the "flash" kernel on Maxwell.
It still uses ingestor credentials to get metadata, but it also has the power to ingest/update metadata inside SciCat.
This way we can do:
- from scicat import ScicatClient
- scicat_ingestor = ScicatClient(proposal_id, source_folder)
- run_info = scicat_ingestor.get_dataset(runNumber)

[...]

PR #347, opened 2024-02-21
token based login for flash scicat

zain-sohail:
Suggested by @kutnyakhov in #345 , simple fix that only affects the flashloader so would be nice to merge quickly.

I will try to add tests, but might be tricky considering it's an API call.

### DIFF
--- a/sed/loader/flash/loader.py
+++ b/sed/loader/flash/loader.py
@@ -750,6 +750,7 @@ class FlashLoader(BaseLoader):
         load_parquet: bool = False,
         save_parquet: bool = False,
         force_recreate: bool = False,
+        **kwds,
     ) -> Tuple[dd.DataFrame, dd.DataFrame]:
         """
         Handles loading and saving of parquet files based on the provided parameters.
@@ -835,7 +836,7 @@ class FlashLoader(BaseLoader):
 
         return dataframe_electron, dataframe_pulse
 
-    def parse_metadata(self, scicat_token: str = None) -> dict:
+    def parse_metadata(self, scicat_token: str = None, **kwds) -> dict:
         """Uses the MetadataRetriever class to fetch metadata from scicat for each run.
 
         Returns:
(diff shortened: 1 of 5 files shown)


zain-sohail:
I have added tests by mocking the requests from server. 

The pytest fixture library for this is requests-mock. 
Tests seems to run but on Pypi it said that it's only valid up to python 3.10 so I am not fully certain if it works on 3.11

rettigl:
I added a manual workflow trigger for the multiversion tests, and ran them on this branch, seems to work fine also with 3.11

rettigl (approved):
LGTM. Good addition would be, as said, a mechanism to provide the token directly from the example notebook in a prepared fashion, so one does not have to edit config files all the time.

sed/loader/flash/loader.py
```diff
@@ -924,7 +925,7 @@ def read_dataframe(
 
         df, df_timed = self.parquet_handler(data_parquet_dir, **kwds)
 
-        metadata = self.parse_metadata() if collect_metadata else {}
+        metadata = self.parse_metadata(**kwds) if collect_metadata else {}
```
rettigl, line 928: Did you test this? You pass the **kwds dict to both parquet_handler and parse_metadata, neither of which accept **kwds arguments. I would expect this to always lead to an unexpected keyword argument exception

zain-sohail: I didn't test the loader function before and you were right. I have fixed this now

rettigl (approved):
I cannot really test this, as I don't have a scicat token, and don't know which URL to use. None if the FLASH metadata stuff is really documented in examples, I suggest this should be still added, so users understand what to do. 
From reading the code, I think it should be fine.

sed/loader/flash/loader.py
```diff
@@ -924,7 +926,7 @@ def read_dataframe(
...
```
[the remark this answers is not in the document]
zain-sohail: Regarding documenting the metadata code in examples, I think it will be done later since scicat is still in test instance.

For **kwds, I don't find a large docstring necessarily useful. For someone going through the API on our docs (or looking at code on IDE), they can see relevant keyword arguments documented for relevant methods.

PR #351, opened 2024-03-04
Fix linting error

### LINKED ISSUE #352: Problem with saving into NeXus format including metadata
Somehow I have a problem to use prc.save("*.nxs").
In particular, it complains with e.g.
"The data entry corresponding to /ENTRY[entry]/INSTRUMENT[instrument]/energy_resolution is required and hasn't been supplied by the reader."
Although this information is provided in the JSON file,
but it became a nested JSON file after last changes with #336 changes.

If I am changing file structure from nested
json_data["/ENTRY[entry]/INSTRUMENT[instrument]"]["energy_resolution"]
to simple JSON structure
json_data["/ENTRY[entry]/INSTRUMENT[instrument/energy_resolution"]
then everything is fine.

@rettigl maybe you can help me with this?

rettigl:
downgrade ruff, and update ruff settings
update pynxtools to pynxtools-mpes

Closes #352


.github/workflows/release.yml
```diff
@@ -11,7 +11,7 @@ on:
     branches:
       - main
     tags:
-      - '*'
+      - v[0-9]+.[0-9]+.[0-9]+
```
zain-sohail, line 14: Does this restrict the versioning to between 0 and 9?
If so, why is it placed? Semantic versioning allows for values above 9 as well. Good example is python 3.11.

zain-sohail (approved):
Only the suggested comment. Otherwise, this can be merged.

.github/workflows/release.yml
rettigl: [0-9]+ is any number of digits, so this should be fine, no? I tested it, however, only for the single-digit case

zain-sohail: I am not knowing much about regex but I looked it up and it seems good

PR #353, opened 2024-03-05
Get metadata from scicat using token

kutnyakhov:
Due to update of SciCat to a new backend at the beginning of the year, not only headers were changed but also URL was corrected (one "/" in front was removed).
Thus, it was possible to get only metadata ingested into SciCat this year using last PR #347 .
This PR now checking if "requests.get(new URL)" returns empty object and then uses instead "requests.get(old URL)".

### DIFF
--- a/sed/loader/flash/metadata.py
+++ b/sed/loader/flash/metadata.py
@@ -89,39 +89,32 @@ class MetadataRetriever:
         """
         headers2 = dict(self.headers)
         headers2["Authorization"] = "Bearer {}".format(self.token)
+        
         try:
-            # Create the dataset URL using the PID
             dataset_response = requests.get(
-                self._create_dataset_url_by_PID(pid),
-                params={"access_token": self.token},
+                self._create_new_dataset_url(pid),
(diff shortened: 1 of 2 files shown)


zain-sohail (approved):
I have tested that data is being loaded so I will merge it now.

rettigl:
> I have tested that data is being loaded so I will merge it now.

Did you check that all example code work without modification? This pr introduces a new config key without changing any config file, nor any appropriate error handling as it appears to me.

zain-sohail:
> > I have tested that data is being loaded so I will merge it now.
> 
> 
> 
> Did you check that all example code work without modification? This pr introduces a new config key without changing any config file, nor any appropriate error handling as it appears to me. 

The key is only necessary if user gives the keyword argument collect_metadata to true. 
Even if it is true, and no relevant config key appears, the code still runs and a warning is sent to user. 
Eitherways this pr just introduces the token based system, because the username and password is depreciated at scicat. So even for old beam times, only this version can work. Hence the urgency.

PR #359, opened 2024-03-15
Fix Jupyter lab version

rettigl:
Fix Jupyter lab to version 3, for compatibility with the desy jhub


zain-sohail:
Could you explain which compatibility it fixes? Do you mean that ipywidgets work then?

zain-sohail (approved):
Tested this to be working

PR #360, opened 2024-03-15
add notebook explaining the effect of jittering

### LINKED ISSUE #180: Add tutorial for correct use of jittering
Add a tutorial notebook illustrating the influence of the jittering parameters, and their correct determination

rettigl:
Closes #180

### DIFF
--- a/tutorial/8_jittering_tutorial.ipynb
+++ b/tutorial/8_jittering_tutorial.ipynb
@@ -1,0 +1,351 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "8ad4167a-e4e7-498d-909a-c04da9f177ed",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "# Correct use of Jittering\n",
+    "This tutorial discusses the background and correct use of the jittering/dithering method implemented in the package."
+   ]
+  },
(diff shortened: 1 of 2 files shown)


tutorial/8_jittering_tutorial.ipynb
```diff
@@ -0,0 +1,351 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "f1f82054",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# The Scan directory\n",
+    "fdir = data_path + '/Scan049_1'\n",
+    "# create sed processor using the config file:\n",
+    "sp = sed.SedProcessor(folder=fdir, config=\"../sed/config/mpes_example_config copy.yaml\")"
```
zain-sohail, line 66: I guess this should be just mpes_example_config

zain-sohail (approved):
Well explained effect of jittering and how to use it. 
Would you say that finding the right parameter is just a brute force method for the user? or are there some reasonable values to use?

Perhaps we can come up with an optimization routine to find the optimal jittering (but that's another discussion)

tutorial/8_jittering_tutorial.ipynb
rettigl: Thanks, fixed

PR #364, opened 2024-03-19
Benchmark

### LINKED ISSUE #362: Add benchmarks for binning and conversion pipeline
Add benchmarks to be run by the update action, and on push to main

rettigl:
Add benchmarks for binning and conversion.

Closes #362

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -4,7 +4,11 @@ name: benchmark
 on:
     workflow_dispatch:
     push:
-        branches: [ main, create-pull-request/patch ]
+        branches: [ main, create-pull-request/patch, benchmark ]
+
+    pull_request:
+      branches:
+        - main
 
 jobs:
   benchmark:
@@ -15,6 +19,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           lfs: true
+
+      - uses: tibdex/github-app-token@v1
+        id: generate-token
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
 
       # Use cached python and dependencies, install poetry
       - name: "Setup Python, Poetry and Dependencies"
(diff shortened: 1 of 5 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,29 @@
+name: benchmark
```
zain-sohail, line 1: Not sure why the very last run of this workflow failed after 5 mins or so

benchmarks/benchmark_sed.py
```diff
@@ -0,0 +1,150 @@
...
+n_pts = 100000000
+ranges = np.array([[0, 2048], [0, 2048], [60000, 120000], [2000, 20000]])
+axes = ["X", "Y", "t", "ADC"]
+array = (
+    dask.array.random.random((n_pts, len(ranges))) * (ranges[:, 1] - ranges[:, 0]) + ranges[:, 0]
+)
+dataframe = dask.dataframe.from_dask_array(array, columns=axes)
+
+
+target_artificial_1d = 2.68
```
zain-sohail, line 30: I'd propose that these values are stored in a yaml file inside this folder. Everytime the values improve, they are updated in that file. So we always have an up to date comparison and hence pass test if performance gets worse than current state.

rettigl: That is a good idea to automatically update targets If they decrease. I can look into this

rettigl: Now with targets in yaml file, and PR mechanism for updates

zain-sohail: Looks good.

From what I understand, a new PR will be created everytime there is a better benchmark (using a factor 1.2, there's some room for runner differences). But doesn't this mean that the github bot will create PR seperately for package updates and benchmark updates? 

I'd rather say that we combine the update dependencies workflow with benchmark workflow (have two jobs like in release workflow for example). Then every week it's updating the packages, benchmarking and opening a PR with the benchmark results. 
If something like the scipy issue occurs, we would be able to address it on the spot. 

It's also fine if you merge this repo for now and we discuss this later.

zain-sohail (approved):

PR #373, opened 2024-03-22
Check for unused arguments and variables

rettigl:
add linter options to check for unused arguments and variables, and add corresponding ignores


zain-sohail (approved):
Good addition.

PR #374, opened 2024-03-23
Fix debugging for parallelized dfield generation

rettigl:
The parallelization using multiprocessing recently introduced into the dfield generation prevents the code to run in debug mode (it just freezes or gives errors). This PR replaces multiprocessing by joblib to allow debugging of the code
Also, it removes the additional loop parallelization, as it almost does not improve performance, and complicates the code substantially.


zain-sohail (approved):
Looks good. There even seems to be an improvement (not sure how reliable) in the benchmark after removal of that extra loop.

PR #386, opened 2024-04-19
Remove conda environvment file

zain-sohail:
This file is 2 years old and developers no longer use env.yml


rettigl (approved):
Strange we did not remove this already

PR #393, opened 2024-05-03
Exception handling for binning when df cols are of non-numeric types

zain-sohail:
Several people have reported a bug per email with 
```
TypingError: Failed in nopython pipeline (step: nopython frontend)
non-precise type array (pyobject, 2d, F)
```
A quick fix was to just use hist_mode: numpy.

Now I finally managed to track it down to the cause. When the dtypes of different cols that are being binned are not the same, such as int16 and float32, numpy treats the array as a type object. 
Usually, numpy infers these types and choose the highest precision type for all in array automatically. However, if you do df.values, it is an object. But somehow choosing columns and then getting values makes numpy choose the array type correctly.

I also added an exception in case the array type is still an object.

image

### DIFF
--- a/sed/binning/binning.py
+++ b/sed/binning/binning.py
@@ -173,8 +173,8 @@ def bin_partition(
         vals = part.iloc[:, col_id].values
     if vals.dtype == "object":
         raise ValueError(
-            "Numba binning requires all binned dataframe columns to be of the same type. "
-            "Encountered data types where "
+            "Binning requires all binned dataframe columns to be of numeric type. "
+            "Encountered data types were "
             f"{[part.columns[id] + ': ' + str(part.iloc[:, id].dtype) for id in col_id]}. "
             "Please make sure all axes data are of numeric type.",
         )
--- a/sed/binning/binning.py
+++ b/sed/binning/binning.py
@@ -171,12 +171,14 @@ def bin_partition(
         vals = sel_part.values
     else:
         vals = part.iloc[:, col_id].values
+    if vals.dtype == "object":
+        raise ValueError(
+            "Numba binning requires all binned dataframe columns to be of the same type. "
+            "Encountered data types where "
+            f"{[part.columns[id] + ': ' + str(part.iloc[:, id].dtype) for id in col_id]}. "
+            "Please make sure all axes data are of numeric type.",
+        )
     if hist_mode == "numba":
-        if vals.dtype == "object":
-            raise ValueError(
-                "Numba binning requires all axes data to be of the same type. "
-                "Please make sure all axes data are of the same type, or use numpy binning. ",
-            )
         hist_partition, edges = numba_histogramdd(
             vals,
             bins=bins,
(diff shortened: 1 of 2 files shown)


zain-sohail:
apparently was addressed here but with a different solution:
https://github.com/OpenCOMPES/sed/blob/702203f9e55128b53be484852650aec1fa443697/sed/core/processor.py#L2021-L2042

rettigl:
I cannot reproduce this problem with the test data. Can you post concrete steps how to reproduce, and maybe add a test for this?

zain-sohail:
> I cannot reproduce this problem with the test data. Can you post concrete steps how to reproduce, and maybe add a test for this?

I will try to reproduce this with test data but this can also be tested with the data/config here:
https://syncandshare.desy.de/index.php/s/58y44ncLFWpggTS

zain-sohail:
This should produce the relevant error:

```python
config_path = "tests/data/loader/flash/config.yaml"
sp = SedProcessor(runs=[43878], config=config_path, collect_metadata=False)

# just as an example
sp._dataframe = sp._dataframe.astype({'dldTimeSteps': 'float32', 'dldPosX': 'uint32', 'dldPosY': 'uint32'})

axes = ['dldTimeSteps',  'dldPosY', 'dldPosX']
bins = (1, 10, 10)
ranges = [[4200, 5000], [460,470], [460,470]]
res = sp.compute(bins=bins, axes=axes, ranges=ranges)
```

rettigl:
The root problem is that your config file does not fit to the data. In your data file, the delayStage channel consists of a 2D array of four different values:
image
This is represented as an "object" type in the dataframe (because each column can only be 1D):
image
So the problem is not in the binning code, but in the dataframe and config. A solution is to add slicing to the config:

```
    # Optical laser channels
    delayStage:
      format: per_train
      group_name: "/FL1/Experiment/Pump probe laser/delay line IK220.0/ENC/"
      slice: 0
```

Then it works with the code from the main branch.

zain-sohail:
> So the problem is not in the binning code, but in the dataframe and config. A solution is to add slicing to the config:

That seems to indeed be the problem. My reasoning for the bug were incorrect. 

However, I still support this change for a few reasons

In case it is incorrectly defined channels by user 
- that are not being binned, this will still bin the correctly defined channels (that are to be binned)
https://github.com/OpenCOMPES/sed/blob/67cc3ec23bc8aa92479f695c035aa9735d1f5fcd/sed/binning/binning.py#L173
- that are being binned, it will throw an error that is clearer than what numba throws (we can change it to let the user know to add the correct slices for example).
https://github.com/OpenCOMPES/sed/blob/67cc3ec23bc8aa92479f695c035aa9735d1f5fcd/sed/binning/binning.py#L175-L179

rettigl (approved):
I checked and this even seems to be a bit faster. 
I adjusted the error message still to be more meaningful, and to perform the check always, as it also does not work with numpy binning, if indeed in the binned columns there are objects left.

PR #395, opened 2024-05-05
Html representation of processor and metadata in notebooks

### LINKED ISSUE #35: Add pretty print for the metadata container (dict)
Add pretty print, possibly to HTML, for the metadata container class. This is for the moment contained as a nested dictionary, which should have a structure as close as possible to the target nexus data structure.

zain-sohail:
This PR introduces an expandable representation of the data and metadata, and perhaps plots. We can discuss what should be included and what not.
Just type `sp` in a notebook after processor class has been instantiated and it will output the html version.

But also the metadata repr is interesting:
To test: you can write `sp.attributes` in a notebook and that will give the html representation.
If `print(sp.attributes)` is used, it gives a yaml representation, which is also more human readable than json.

closes #35

### DIFF
--- a/sed/core/metadata.py
+++ b/sed/core/metadata.py
@@ -44,6 +44,14 @@ class MetaHandler:
         html = self._format_attributes(self._m)
         return html
 
+    @property
+    def metadata(self) -> Dict:
+        """Property returning the metadata dict.
+        Returns:
+            dict: Dictionary of metadata.
+        """
+        return self._m
+
     def add(
         self,
         entry: Any,
--- a/sed/core/metadata.py
+++ b/sed/core/metadata.py
@@ -1,5 +1,4 @@
 """This is a metadata handler class from the sed package
-
 """
 import json
 from copy import deepcopy
@@ -10,7 +9,8 @@ from sed.core.config import complete_dictionary
 
 
 class MetaHandler:
-    """[summary]"""
+    """This class provides methods to manipulate metadata dictionaries,
+    and give a nice representation of them."""
 
     def __init__(self, meta: Dict = None) -> None:
         self._m = deepcopy(meta) if meta is not None else {}
@@ -22,13 +22,14 @@ class MetaHandler:
         return json.dumps(self._m, default=str, indent=4)
 
     def _format_attributes(self, attributes, indent=0):
+        INDENT_FACTOR = 20
         html = ""
         for key, value in attributes.items():
             # Format key
             formatted_key = key.replace("_", " ").title()
             formatted_key = f"<b>{formatted_key}</b>"
 
-            html += f"<div style='padding-left: {indent * 10}px;'>"
+            html += f"<div style='padding-left: {indent * INDENT_FACTOR}px;'>"
             if isinstance(value, dict):
                 html += f"<details><summary>{formatted_key} [{key}]</summary>"
                 html += self._format_attributes(value, indent + 1)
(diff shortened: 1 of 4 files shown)


zain-sohail:
Similar additions were attempted in #305 but those are more specific to loaders with runs and maybe to Flash. We can see how to include those.

sed/core/processor.py
```diff
@@ -178,10 +178,35 @@ def __repr__(self):
...
-        pretty_str = df_str + "\n" + attributes_str
+        pretty_str = df_str + "\n" + "Metadata: " + "\n" + self._attributes.__repr__()
         return pretty_str
 
+    def _repr_html_(self):
+        html = "<div>"
+
+        html += (
+            f"<details><summary>Dataframe</summary>{self.dataframe.head()._repr_html_()}</details>"
+        )
```
rettigl, line 189: Computing the head sometimes takes a substantial amount of time (up to 30 s with the tutorial2). I would not do this per default, as the repr should be fast, I would say.

sed/core/processor.py
```diff
@@ -178,10 +178,35 @@ def __repr__(self):
...
+
+        # Add expandable section for dataframe
+        html += f"<details><summary>Dask</summary>{self.dataframe._repr_html_()}</details>"
+
```
rettigl, line 193: I somehow find this redundant information to the first part. I'd suggest to remove the "Dataframe" section, and move this code there.

sed/core/processor.py
```diff
@@ -178,10 +178,35 @@ def __repr__(self):
...
+        html += "<details><summary>Metadata</summary>"
+        html += self.attributes._repr_html_()
+        html += "</details>"
+
+        # Add expandable section for plots
+        html += "<details><summary>Plots</summary>"
+        # Something like the event histogram can be added here,
+        # but the method needs to output image/html
+        # self.view_event_histogram(dfpid=2, backend="matplotlib")
+        html += "</details>"
```
rettigl, line 204: As this should be fast, again I am not sure how helpful this is, in addition as there is the separate method for it. Wen can discuss, though.

rettigl (commented):
I generally like this a lot, however computing the head takes quite some time apparently, making the output quite slow.
Maybe we should discuss which parts of the beamtime addition to the representation we want to add.
Also, we should have tests for this as well (to see that it runs correctly for the different loaders, without a dataframe loaded, etc.).

rettigl:
Part of the long processing time apparently comes from the metadata conversion:
grafik
This is probably due to large arrays being converted into html, so we should maybe exclude these and replace them with an information about the array shape or so.

rettigl:
I'd also suggest to move the first layer of metadata one inset in already:
grafik
Because the expandable elements then look like first-layer elements. Maybe generally for all entries (also the dataframe etc.).

rettigl:
> Part of the long processing time apparently comes from the metadata conversion: grafik This is probably due to large arrays being converted into html, so we should maybe exclude these and replace them with an information about the array shape or so.

It mostly is due to the large deformation fields:
grafik

sed/core/metadata.py
```diff
@@ -18,8 +20,31 @@ def __getitem__(self, val: Any) -> None:
...
-        # TODO: #35 add pretty print, possibly to HTML
-        return str(self._m)
+        return yaml.dump(self._m, allow_unicode=True, default_flow_style=False)
+
+    def _format_attributes(self, attributes, indent=0):
+        html = ""
+        for key, value in attributes.items():
+            # Format key
+            formatted_key = key.replace("_", " ").title()
+            formatted_key = f"<b>{formatted_key}</b>"
```
rettigl, line 30: I'd suggest to remove this, or at least keep the real key in brackets or something. It makes parsing the dictionary based on its representation impossible or very difficult.

sed/core/processor.py
zain-sohail: I have created another method called overview_panel in comments but can be another issue/PR.

zain-sohail:
> Part of the long processing time apparently comes from the metadata conversion: grafik This is probably due to large arrays being converted into html, so we should maybe exclude these and replace them with an information about the array shape or so.

Seems to not be the case. Even with shape, it is slow. 

I tried to check directly with a test class and it outputs the html in 0 s
image

I am trying to figure out where the slowdown is.

sed/core/metadata.py
zain-sohail: I have put the key in [] now.

sed/core/processor.py
zain-sohail: Good point. Was just an initial idea to choose between dask or head.

zain-sohail:
The problem has been resolved. It was the `__repr__` dumping to yaml that took so long. And somehow when `_repr_html` is called, it also calls `__repr__`

zain-sohail:
> we should have tests for this as well (to see that it runs correctly for the different loaders, without a dataframe loaded, etc.).

I have added tests to the Metadata class and it's repr/repr_html but not sure how to test what you suggested.

rettigl (approved):
LGTM now. 
We can still add part of Steinn's code, or do this in a later PR (add issue then).

Issue #249, opened 2023-11-07
Implement Logging instead of prints


steinnymir:
It was suggested in #177 to implement logging for verbose output of all functionalities.

That issue was closed as the requested changes were made, but this implementation request still stands

rettigl:
Addressed in PR #490

rettigl:
Completed via #490

PR #401, opened 2024-05-14
Module for datasets in SED

### LINKED ISSUE #403: Use standardized directories
To promote sed as a proper application, it's ideal if the user config files/data/logs be saved in OS specific standardized directories. 

`platformdirs` provides such functionality and can be used to save user configs and other data, if user doesn't specify where to put them. 
Logs can also be saved in the user log directory so when there are bugs, user can just provide us the log file.

zain-sohail:
This PR introduces:

- Easy API for getting and removing datasets `[Stored in ./datasets/<data_name>/]`
- A Datasets config file (JSON) to add new datasets (or remove) `[Folder config stored in ./datasets.json]`
  - Herirachically merges the module, user and folder files into one completed datasets.json 
- Introduces logging setup for this module `[Now stored in current working dir ./logs]`
- Using platformdirs to store user state of datasets 
  - datasets.json is copied to user_config_path and updated there to keep track of state
- Tutorials are updated to use the new dataset module these changes

These changes are arranged in two classes. 

- to get/remove datasets: **`Dataset`** which user only accesses the class instance **`dataset`**
`from sed.dataset import dataset`
- to configure (url etc.) for new/old datasets: **`DatasetsManager`**
`from sed.dataset import DatasetsManager`

Please look at this notebook for complete usecase: [datasets_example.ipynb.zip](https://github.com/user-attachments/files/15559614/datasets_example.ipynb.zip)

 related to #249 and closes #403


zain-sohail:
closes #249 and closes #403

rettigl:
Did not look in detail, but documentation is no longer working, so please fix this first.

sed/core/logging.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def setup_logging(name: str, verbose: bool = False, debug: bool = False) -> logging.Logger:
+    """
+    Configures and returns a logger with specified log levels for console and file handlers.
+
+    Args:
+        name (str): The name of the logger.
+        verbose (bool): If True, sets the console log level to INFO. Defaults to False.
+        debug (bool): If True, sets the console log level to DEBUG. Defaults to False.
```
rettigl, line 19: Why not just use one parameter "log_level"?

sed/core/logging.py
```diff
@@ -0,0 +1,67 @@
...
+    console_handler.setLevel(console_level)
+
+    # Create formatter for console
+    console_formatter = logging.Formatter("%(message)s")
+    console_handler.setFormatter(console_formatter)
+
+    # Add console handler to logger
+    logger.addHandler(console_handler)
+
+    log_file = USER_LOG_PATH.joinpath("sed.log")
```
rettigl, line 51: Shouldn't the file names contain e.g. a date or so?

sed/core/user_dirs.py
```diff
@@ -0,0 +1,45 @@
...
+
+from pathlib import Path
+
+from platformdirs import user_config_path
+from platformdirs import user_data_path
+from platformdirs import user_log_path
+
+USER_CONFIG_PATH = user_config_path(appname="sed", appauthor="OpenCOMPES", ensure_exists=True)
+USER_LOG_PATH = user_log_path(appname="sed", appauthor="OpenCOMPES", ensure_exists=True)
+USER_DATA_PATH = user_data_path(appname="sed", appauthor="OpenCOMPES", ensure_exists=True)
```
rettigl, line 20: How can this be configured?

sed/dataset/dataset.py
```diff
@@ -0,0 +1,285 @@
...
+    If the file does not exist, it copies the default datasets JSON file from the module
+    directory to the user's datasets directory.
+
+    Returns:
+        dict: The datasets dict loaded from user's datasets JSON file.
+    """
+    # check if datasets.json exists in user_config_dir
+    if not os.path.exists(json_path_user):
+        shutil.copy(json_path_module, json_path_user)
+    datasets = load_config(str(json_path_user))
```
rettigl, line 38: With this, the module version is ignored if a user version exists. Why not use the same functionatily as for the config?

sed/dataset/dataset.py
```diff
@@ -0,0 +1,285 @@
...
+
+def available_datasets() -> list:
+    """
+    Returns a list of available datasets.
+
+    Returns:
+        list: List of available datasets.
+    """
+    datasets = load_datasets_dict()
+    return list(datasets)
```
rettigl, line 50: Wouldn't it make sense to put all this into a class, and make this an attribute?

sed/dataset/dataset.py
```diff
@@ -0,0 +1,285 @@
...
+
+    Returns:
+        str: The final data path for the dataset.
+    """
+    # Notify the user if data might already exist
+    if existing_data_path and data_path and existing_data_path != data_path:
+        logger.warning(
+            f'{data_name} data might already exists at "{existing_data_path}", '
+            "unless deleted manually.",
+        )
```
rettigl, line 94: Why is this a warning?

sed/dataset/dataset.py
rettigl, line 1: I don't really see the reason for this code. You essentially create 285 lines to do what ~4 lines in each tutorial do...
What functionality does this add, that was not there before?

rettigl (commented):
Ì'm not really sure I understand the purpose of this dataset module. How do you intend it to be used? What is the advantage over the current lines of code (you add ~400 or so lines of code, without a clear purpose or use case that is apparent to me).
The logging part is probably nice, but I suggest to separate this out into another PR. We shall have to discuss which of the default outputs can be put into logs, I am not sure.

zain-sohail:
> Ì'm not really sure I understand the purpose of this dataset module. How do you intend it to be used? What is the advantage over the current lines of code (you add ~400 or so lines of code, without a clear purpose or use case that is apparent to me). 

I understand your point. The lines of code seem a lot due to a lot done for documentation/testing. There are a few reasons why I decided to work on this:

- Most user and data focused repositories have functionality to get datasets for the user. It's one thing for user to run the tutorials, but another if they want to just fetch the given datasets in their own pipelines. 
- New datasets can easily be added and used. We just need to add to the json file (I thought of adding an entry point for user so they can add other data from zenodo etc. to their datasets.json).
- User keeps track of data they downloaded and doesn't redownload/reextract them. I faced this problem because the notebooks/scripts can exist in different places and use relative paths. With having access to state (from datasets.json in user config), the user can get to know where the data should already exist. Moreover, with the flash data, where the data needs to be rearranged (likely case for most data), I once deleted the empty directory and that made it download the whole file again. 

In short the additional complexity in the package allows for simplicity for user.

You to check how datasets.json in your config dir is updated to see the benefit e.g.

- Use `load_dataset(<dataset_name>)` to get the data
- Use `load_dataset(<dataset_name>, <own_path>)` to get the data
- Try to reload the same dataset with another path `load_dataset(<dataset_name>, <anonther_path>)`

> The logging part is probably nice, but I suggest to separate this out into another PR. We shall have to discuss which of the default outputs can be put into logs, I am not sure.

I can do that. But I put it here because the use case was quite apparent (good test case)
Currently log.warning and above is shown to user and everything else just written to the log file.

sed/core/logging.py
zain-sohail: I wanted to do that but since SedProcessor currently has a bool flag for verbose I tried to keep the compatibility. We can definitely change that.

sed/core/logging.py
zain-sohail: Currently everything is logged into one file but your idea makes sense. Should be an easy addition. I can either update it here, split to another PR or merge this PR and update logging afterwards in new PR.

sed/core/user_dirs.py
zain-sohail: platformdirs basically gives you the standard paths defined by the OS where config, logs and data should go. By giving the appname (we could name it sed-processor) and appauthor, it creates these directories for us and we just have to use it for our application.
These don't need to be configured. Or maybe I misunderstood your question.

sed/dataset/dataset.py
zain-sohail: Yes. The advantage is, the user doesn't need to look for that file. It is where it is expected to be.
Or do you mean the recursive lookup? That I didn't think about but can be added.

sed/dataset/dataset.py
zain-sohail: Yes. So you suggest the user first creates a class instance and then call class.load_dataset() etc?

sed/dataset/dataset.py
zain-sohail: Since log level for user is set to warning at default and I think this is important information to know. Otherwise you download the dataset again for no reason.

sed/core/logging.py
rettigl: Okay, we can change this then once we integrate this logging into the other modules

sed/core/logging.py
rettigl: Apart from splitting by date, I would also strongly suggest to keep the log files together with the notebooks. If you put them into a global log file, it's going to be impossible to assign which lines come from which notebook, run etc.

sed/core/user_dirs.py
rettigl: I think it's not a good idea to put large amounts of data into a folder the user does not have any control over. This will lead to overflowing root drives, huge amounts of data being synchronized, etc.

sed/dataset/dataset.py
rettigl: This way, any changes to the package (added data sets) will not be available for a user who has used this in an old version, because his local copy does not contain these then. If you use the hierarchical dict merging of the config module, you can avoid this.

sed/dataset/dataset.py
rettigl: It's true this will make it somewhat more difficult to use. Up for discussion.

sed/dataset/dataset.py
rettigl: Maybe I don't understand the logic well here. When is this warning supposed to be issued? How is existing_data_path determined by the caller, i.e. how do you find all possible existing versions of this data set?

rettigl:
> * Most user and data focused repositories have functionality to get datasets for the user. It's one thing for user to run the tutorials, but another if they want to just fetch the given datasets in their own pipelines.
> 

I see this point, but then we should see how it is done for other package large scale example data (e.g. xarray, pandas, etc.). typically you can just import datasets from a module or so.

> * New datasets can easily be added and used. We just need to add to the json file (I thought of adding an entry point for user so they can add other data from zenodo etc. to their datasets.json).
> 

For some example data this might be helpful, but for the general use case for converting data from the lab/beamline, this won't happen, no?

> * User keeps track of data they downloaded and doesn't redownload/reextract them. I faced this problem because the notebooks/scripts can exist in different places and use relative paths. With having access to state (from datasets.json in user config), the user can get to know where the data should already exist. Moreover, with the flash data, where the data needs to be rearranged (likely case for most data), I once deleted the empty directory and that made it download the whole file again.

This indeed is true. I need to check this.

> I can do that. But I put it here because the use case was quite apparent (good test case)
> Currently log.warning and above is shown to user and everything else just written to the log file.

We can use this here as an example then, and integrate it for the other modules. Then this PR however does not close #249

sed/core/user_dirs.py
zain-sohail: I understand your point. Storing to user data will not be default behavior then.

sed/dataset/dataset.py
zain-sohail: I will update to suggestion.

sed/dataset/dataset.py
zain-sohail: Currently, only one path is saved to existing version. When data is downloaded, extracted (and rearranged), that path is saved to the user config. The file list is also saved and can be compared to see if data is corrupted (maybe by user deleting something) or is how it is supposed to be.
I should extend this to include not just one path but multiple paths the data is downloaded to.

zain-sohail:
> I see this point, but then we should see how it is done for other package large scale example data (e.g. xarray, pandas, etc.). typically you can just import datasets from a module or so.

Looking here, that's true:
https://scikit-learn.org/stable/datasets/toy_dataset.html
due to size being small.

But for example this downloads the dataset first. I will look how they do their data loading procedure.
https://pytorch.org/tutorials/beginner/basics/data_tutorial.html 
Another example with source: https://github.com/keras-team/keras/tree/v3.3.3/keras/src/datasets

> For some example data this might be helpful, but for the general use case for converting data from the lab/beamline, this won't happen, no?

At the end, this just downloads/extracts/arranges the data on the specified path. One could also do this with lab/beamtimes if a url for that set is given. But everything afterwards happens with the SedProcessor.
The example I imagine is when I want to look at other interesting single event datasets from zenodo, for arpes or maybe even other single event applications.

> We can use this here as an example then, and integrate it for the other modules. Then this PR however does not close #249
Noted.

> This indeed is true. I need to check this.

Improvements can definitely be made to the current state.

- [ ] Account for multiple paths where same dataset can exist
- [ ] Take snapshot of file state after download, after extraction and after arrangment so those states are saved more robustly (right now only the final file state is captured)
- [ ] Remove default user data path. User has to either define their own path, or specifically ask to use `default` 
- [ ] Use heirarchical dict merging already in use by config files
- [ ] Maybe put this all in a class
- [ ] add option to delete datasets
- [ ] log files with notebooks
- [ ] let user add more datasets

rettigl (commented):
I like this latest version in most parts. Only major Issue I have is that the current way of unpacking the files does not conserve the creation dates and times from the archives, which produces wrong time stamps for the data from our lab. This produces wrong results e.g. for tutorial 6.
I would also find it helpful to have the tutorial for the dataset manager included in the docs.

zain-sohail:
> I like this latest version in most parts. Only major Issue I have is that the current way of unpacking the files does not conserve the creation dates and times from the archives, which produces wrong time stamps for the data from our lab. This produces wrong results e.g. for tutorial 6. I would also find it helpful to have the tutorial for the dataset manager included in the docs.

I have updated this now to use the correct timestamps.
I looked into just using the linux unzip but that's not OS agnostic, it seems. Should work on Linux and Mac but not on Windows.
But this change should work now.

The only thing remaining is the docs build. I have looked into caching the docs build and but will make a seperate PR for that. For this, I will just try to make docs work ad hoc.

zain-sohail:
One thing I found on stack overflow. It is possible to get zip files and not save to disk but directly extract them. But I propose a seperate issue for that if necessary.

rettigl:
> I have updated this now to use the correct timestamps. I looked into just using the linux unzip but that's not OS agnostic, it seems. Should work on Linux and Mac but not on Windows. But this change should work now.

Yes, at least on linux this works now.

> 
> The only thing remaining is the docs build. I have looked into caching the docs build and but will make a seperate PR for that. For this, I will just try to make docs work ad hoc.

Why can't you use the same mechanism as before for docs build? The download time is just a fraction of the compile time, so I don't see to much reason for caching...

zain-sohail:
> > I have updated this now to use the correct timestamps. I looked into just using the linux unzip but that's not OS agnostic, it seems. Should work on Linux and Mac but not on Windows. But this change should work now.
> 
> Yes, at least on linux this works now.

I meant if using the linux cli tools. But python's module is os agnostic.


> Why can't you use the same mechanism as before for docs build? The download time is just a fraction of the compile time, so I don't see to much reason for caching... Anyways, free cache size is much too small, no?

I meant to cache the docs compilation (50 MB usually). If the _build folder is available to sphinx, it updates only the changed files. This makes the build extremely fast. Generally it's simple but due to the notebooks, it requires more finesse.

rettigl:
> I meant to cache the docs compilation (50 MB usually). If the _build folder is available to sphinx, it updates only the changed files. This makes the build extremely fast. Generally it's simple but due to the notebooks, it requires more finesse.

I don't really understand the benefit. The thing that takes the most time is the compilation of the notebooks. The rest takes maybe 10-20s or so. I don't think it is worth building a complicated caching system to safe <1% of the time. And the notebook building you have to do anyways, otherwise you don't know if something changed in them...

zain-sohail:
The docs building now works:
https://github.com/OpenCOMPES/sed/actions/runs/9507832092

I have also included the tutorial notebook I provided along with API/config doc. It will be under dataset after building.



> I don't really understand the benefit. The thing that takes the most time is the compilation of the notebooks. The rest takes maybe 10-20s or so. I don't think it is worth building a complicated caching system to safe <1% of the time. And the notebook building you have to do anyways, otherwise you don't know if something changed in them...

Regarding this, I will make a new issue but this is not urgent problem. 
But in essence, if nbsphinx (that compiles the notebooks) sees that our notebooks haven't changed, it will not compile them. And the only way that works is if we have both the tutorial notebooks (because currently we copy the notebooks from tutorial to docs/tutorial) along with _build folder are cached, since that retains the timestamps and state for nbsphinx to understand. So after using the cache, we only copy the tutorials that have changed (can use a hash function for that) so nbsphinx realizes that those are updated notebooks.
If somehow, we don't have to copy the notebooks, I believe we can get by with caching _build only.

rettigl:
> But in essence, if nbsphinx (that compiles the notebooks) sees that our notebooks haven't changed, it will not compile them.

That is not sufficient, as even if the notebooks themselves did not change, the output of the code might. So you need to process then in order to know.

zain-sohail:
> That is not sufficient, as even if the notebooks themselves did not change, the output of the code might. So you need to process then in order to know.

That's a good point. I guess there's no way to really speed it up then.

rettigl (approved):
First this did not work for me, because my old user datasets.json was out of date. By deleting this, it now works.
I did not check everything now again, but as it appears to run, LGTM.
I added a couple of things:
- gitignore for datasets
- use tqdm.auto
- remove existing logging handlers when initializing loggers, to prevent multiple same loggers if files are being reloaded during a session

rettigl:
Only thing I find still strange is that you get the warning everytime that a dataset is already there. It appears strange, because there is nothing wrong or to warn or even info about, when a dataset is already there. The using existing data message should be enought, for my taste.

rettigl:
I also found it helpful to also print the log-level on the console

rettigl:
Another thing I fixed now is that the rearrange action will overwrite existing files (if they are there, but missing in the config). It gave an error before.

zain-sohail:
> First this did not work for me, because my old user datasets.json was out of date. By deleting this, it now works. I did not check everything now again, but as it appears to run, LGTM. I added a couple of things:

Yes this was a problem for me too. But I think that's a problem mostly during development. The only time I can imagine it maybe failing is when user deletes the dataset not through API but by themselves.
But if that becomes an issue, we can update the code somehow.

Thanks for the updates/suggestions. I updated it to them.

PR #411, opened 2024-06-02
Energy calibration bias shift

### LINKED ISSUE #138: Unify axis naming and definition
Unify beween mps and FLASH

### LINKED ISSUE #399: Offset energy calibration by sample bias
Make energy calibration work independent of sample bias (remove offset)

rettigl:
Shift energy calibration by bias voltages, and apply shift of dataset bias upon adding energy column.
Unify dataframe channels config

closes #399
closes #138


zain-sohail:
I understand that you want to merge to v1 feature branch because these are all breaking changes?

rettigl:
> I understand that you want to merge to v1 feature branch because these are all breaking changes?

Yes, they reqire configuration changes

tutorial/2_conversion_pipeline_for_example_time-resolved_ARPES_data.ipynb
```diff
@@ -472,12 +472,10 @@
    "source": [
     "# use the refid of the bias that the measurement was taken at\n",
```
zain-sohail, line 473: refid isnt used anymore

sed/config/mpes_example_config.yaml
```diff
@@ -79,6 +73,25 @@ dataframe:
...
+      format: per_electron
+      group_name: "/Stream_1"
+    t:
+      format: per_electron
+      group_name: "/Stream_2"
+    ADC:
+      format: per_electron
+      group_name: "/Stream_4"
+    sampleBias:
+      format: per_file
```
zain-sohail, line 92: is this a slow channel format? perhaps we should also change the per_train channels to per_file if it's the same

sed/config/mpes_example_config.yaml
```diff
@@ -79,6 +73,25 @@ dataframe:
     kx: '1/A'
     ky: '1/A'
 
+  # dataframe channels and group names to read from the h5 files
+  channels:
+    # The timestamp
```
zain-sohail, line 78: i understand your timestamps come from the msmarkers group so what does this comment mean here?
Also, the ms_markers_group doesn't seem to be defined in the config but is hardcoded as default values?

sed/loader/mpes/loader.py
```diff
@@ -42,12 +42,9 @@ def hdf5_to_dataframe(
...
     Args:
         files (List[str]): A list of the file paths to load.
-        group_names (List[str], optional): hdf5 group names to load. Defaults to load
-            all groups containing "Stream"
-        alias_dict (Dict[str, str], optional): Dictionary of aliases for the dataframe
-            columns. Keys are the hdf5 groupnames, and values the aliases. If an alias
-            is not found, its group name is used. Defaults to read the attribute
-            "Name" from each group.
+        channels (Dict[str, str], optional): hdf5 channels names to load. Each entry in the dict
+            should contain the keys "format" and "groupName". Defaults to load all groups
```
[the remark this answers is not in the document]
rettigl: thanks, fixed.

sed/config/mpes_example_config.yaml
```diff
@@ -79,6 +71,25 @@ dataframe:
...
+    X:
+      format: per_electron
+      group_name: "/Stream_0"
```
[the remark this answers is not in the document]
rettigl: Indeed, they are datasets and not groups. For sxp, we currently have dataset_key, and index_key. So maybe best to just adopt dataset_key here.

sed/config/mpes_example_config.yaml
```diff
@@ -79,6 +71,25 @@ dataframe:
...
+    Y:
+      format: per_electron
+      group_name: "/Stream_1"
```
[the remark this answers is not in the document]
rettigl: They are supported, see here: 
https://github.com/OpenCOMPES/sed/blob/d0fefd9550f262d51a92ffc90acb01816745654a/sed/loader/mpes/loader.py#L295

sed/loader/mpes/loader.py
```diff
@@ -254,18 +241,39 @@ def hdf5_to_array(
...
     # Delayed array for loading an HDF5 file of reasonable size (e.g. < 1GB)
 
+    # determine group length from per_electron column:
+    nelectrons = 0
+    for channel in channels:
+        if channel["format"] == "per_electron":
+            nelectrons = len(h5file[channel["group_name"]])
+            break
+    if nelectrons == 0:
+        raise ValueError("No 'per_electron' columns defined, or no hits found in file.")
```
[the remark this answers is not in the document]
rettigl: I added this, because I need to no the number of events in the file, to generate the per_file columns. I think empty files will also break the time_stamp functionality anyways, so they are currently not supported.

tutorial/2_conversion_pipeline_for_example_time-resolved_ARPES_data.ipynb
rettigl: fixed

sed/config/mpes_example_config.yaml
rettigl: Yes, these are stored once per file (about 10 Mevents). Currently, this performs very poorly.

sed/config/mpes_example_config.yaml
rettigl: This is a copy paste relict, fixed. 
The ms_marker_group is defined outside the channels, because it does need a separate mechanism. Maybe I also change this, but this also performs rather poorly, so I want to keep the option of only generating them if necessary.

rettigl:
This code here unfortunately performs rather poorly. The generation of the per-file data columns alone make the computation ~20% or so slower, adding the energy offset per electron multiplies the binning time by a factor of almost 4!

sed/config/mpes_example_config.yaml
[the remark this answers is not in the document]
zain-sohail: Yes, I would suggest that. Then I will also update the flashloader to only use index_key and dataset_key and we have a consistent set.

sed/config/mpes_example_config.yaml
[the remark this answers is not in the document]
zain-sohail: I see. I was using the variable `dtype`. See https://github.com/OpenCOMPES/sed/blob/8dd5e6ad2e51f1cf9c6e59083f8233ad26787ac9/sed/loader/flash/buffer_handler.py#L200-L204
but whatever schema we decide is fine.

sed/config/mpes_example_config.yaml
zain-sohail: I think the filling operations are always slow so there would be a reduction in performance. 
Also, per_train channels aren't necessarily stored per_file so I guess we need that

sed/config/mpes_example_config.yaml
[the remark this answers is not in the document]
rettigl: I changed to dtype now as well.

tutorial/2_conversion_pipeline_for_example_time-resolved_ARPES_data.ipynb
```diff
@@ -470,14 +470,11 @@
    "metadata": {},
```
zain-sohail, line 470: The comment above should remove "Here, one can also define a reference id" since I saw that it's no longer used.

zain-sohail (changes requested):
I have tried to review but currently, I don't understand the internals of the calibration methods enough to  review this properly. I have tried because I wanted to adapt these for the side band correction but failed to understand.
Honestly, It'd be great if a small guide on the internals can be made. We can then also put it in user documentation.
If that's too much, if you could explain why you made these changes, I should be able to review based on that.

The only thing I noticed currently was that the tutorials still have the ref_id (in methods find_bias_peaks etc.) whereas the code only uses the ref_energy now.

tutorial/2_conversion_pipeline_for_example_time-resolved_ARPES_data.ipynb
rettigl: Thanks, this has been corrected.

rettigl:
> I have tried to review but currently, I don't understand the internals of the calibration methods enough to review this properly. I have tried because I wanted to adapt these for the side band correction but failed to understand. Honestly, It'd be great if a small guide on the internals can be made. We can then also put it in user documentation. If that's too much, if you could explain why you made these changes, I should be able to review based on that.
> 
> The only thing I noticed currently was that the tutorials still have the ref_id (in methods find_bias_peaks etc.) whereas the code only uses the ref_energy now.

Thank you for your effort. I will for now comment on the changes to explain them better. If I find time, I can also make a more general documentation item out of this.

sed/calibrator/energy.py
```diff
@@ -2200,22 +2211,20 @@ def residual(pars, time, data, binwidth, binning, energy_scale):
     ecalibdict["t0"] = result.params["t0"].value
     ecalibdict["E0"] = result.params["E0"].value
     ecalibdict["energy_scale"] = energy_scale
+    energy_offset = pfunc(-1 * ref_energy, pos[0])
+    ecalibdict["E0"] = -(energy_offset - vals[0])
```
rettigl, line 2215: We build a partial function with offset and TOF as free parameters, and evaluate this at the TOF value (pos[0]) of the reference feature in the first trace, with given energy value (-ref_energy). This returns the difference of the calibration method to the expected value + the first bias voltage. This value minus the respective bias voltage is the negative offset to apply to obtain the correct energy value.

sed/calibrator/energy.py
```diff
@@ -2200,22 +2211,20 @@ def residual(pars, time, data, binwidth, binning, energy_scale):
...
 
-    if (ref_energy is not None) and (t is not None):
-        energy_offset = pfunc(-1 * ref_energy, pos[ref_id])
-        ecalibdict["axis"] = pfunc(-energy_offset, t)
-        ecalibdict["E0"] = -energy_offset
-        ecalibdict["refid"] = ref_id
+    if t is not None:
+        ecalibdict["axis"] = pfunc(ecalibdict["E0"], t)
```
rettigl, line 2218: Here we can now calculate the energy axis (for the first bias value).

sed/calibrator/energy.py
```diff
@@ -2312,12 +2313,10 @@ def poly_energy_calibration(
     ecalibdict["Tmat"] = t_mat
     ecalibdict["bvec"] = bvec
     ecalibdict["energy_scale"] = energy_scale
+    ecalibdict["E0"] = -(pfunc(-1 * ref_energy, pos[0]) + vals[0])
```
rettigl, line 2316: Similar considerations here.

sed/core/dfops.py
```diff
@@ -392,53 +392,20 @@ def offset_by_other_columns(
...
-    df[target_column] = df.map_partitions(
-        shift_by_mean,
-        cols=[col for col, red in zip(offset_columns, reductions) if red],
-        signs=signs_dict,
-        means=means,
-        meta=df[target_column].dtype,
-    )
+    for col, red in zip(offset_columns, reductions):
+        if red == "mean":
+            df[target_column] = df[target_column] + signs_dict[col] * df[col].mean()
```
rettigl, line 398: The direct dataframe operations are substantially faster than map_partition according to my tests. Some of the map_partition calls also fail with newer dask versions.

sed/loader/mpes/loader.py
rettigl, line 1: Changes here are to change the config style, and to allow per-file channels, e.g. for the bias voltage to be added to the dataframe.

rettigl:
For using the calibration function for features from sidebands, do the following:
1. Extract the TOF values of the sidebands (e.g. TOF=[t0, t1, t2, t3, t4, t5, t6]).
2. Generate the corresponding energies, shifted by the bias voltage of your measurement, e.g. energies=[-3hv, -2hv, -hv, 0, hv, 2hv, 3hv] - (maybe +) bias
3. Calculate the energy of the first side band, e.g. E_corelevel - 3hv
4. Provide these values to ec.calibrate(ref_energy=reference, landmarks=TOF, biases=energies, ...)
This should work, I think.

sed/calibrator/energy.py
```diff
@@ -877,6 +881,20 @@ def append_energy_axis(
...
+            df[energy_column] = df[energy_column] + scale_sign * bias_voltage
+        elif self._config["dataframe"]["bias_column"] in df.columns:
+            df = dfops.offset_by_other_columns(
+                df=df,
+                target_column=energy_column,
+                offset_columns=self._config["dataframe"]["bias_column"],
+                weights=scale_sign,
+            )
+        else:
+            print("Sample bias data not found or provided. Calibrated energy might be incorrect.")
```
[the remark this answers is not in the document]
zain-sohail: Perhaps better to throw an error as this is not intended behavior?

sed/calibrator/energy.py
```diff
@@ -1625,10 +1649,7 @@ def add_offsets(
         if constant:
             if not isinstance(constant, (int, float, np.integer, np.floating)):
                 raise TypeError(f"Invalid type for constant: {type(constant)}")
-            df[energy_column] = df.map_partitions(
-                lambda x: x[energy_column] + constant,
-                meta=(energy_column, np.float64),
-            )
+            df[energy_column] = df[energy_column] + constant
```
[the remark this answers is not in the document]
zain-sohail: From my experience, this also works fine. Dask also uses normal pandas workflows in their docs
See `df = df[df.name == 'Alice'] ` at https://docs.dask.org/en/stable/dataframe-best-practices.html

sed/calibrator/energy.py
zain-sohail: I'd suggest to rename this partial function to also tof2ev as it will just use the local scope. 
Using pfunc is not fully explanatory.

sed/calibrator/energy.py
```diff
@@ -2082,8 +2103,7 @@ def fit_energy_calibration(
     vals: list[float] | np.ndarray,
```
zain-sohail, line 2103: the vals variable and other such can have more descriptive names so one doesn't need to go through the docstrings to understand

sed/calibrator/energy.py
zain-sohail: don't really see a need for a partial function here when it's only one argument. But I understand if it's there for a reason

zain-sohail (approved):
Just a few minor comments. Thanks for explaining! Helped a lot. Could be a start to the calibration guide later.

sed/calibrator/energy.py
rettigl: I would not add more changes to the existing code now. I can open an issue. As for the first point, if the bias voltage is not available, you still get a "correct" energy axis, but offset by the bias voltage applied. So I would not consider this a critical failure, but something to notify about.

sed/calibrator/energy.py
rettigl: Added as issue

sed/calibrator/energy.py
rettigl: I'll look into this eventually. It's some old code.

sed/calibrator/energy.py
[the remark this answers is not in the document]
rettigl: As mentioned below, I don't think this is a critical error. It still provides a valid calibration, just with an energy offset. Otherwise, no calibration is possible if you don't know the bias.

PR #415, opened 2024-06-07
Optional dependencies

zain-sohail:
I noticed that while some notebook dependencies were tagged optional, others weren't. And the pip users wouldn't be installing the optional ones without doing 
`pip install sed-processor[notebook]`

So I have made all notebook dependencies optional. User can install with 
`pip install sed-processor[all]`
or 
`pip install sed-processor[notebook]`
currently, since there are no other optional dependencies groups.

This is reflected in the README now.


pyproject.toml
```diff
@@ -40,14 +37,18 @@ tqdm = ">=4.62.3"
 xarray = ">=0.20.2"
 joblib = ">=1.2.0"
 pyarrow = ">=14.0.1"
+fastparquet = ">=0.8.0"
+ipympl = {version = ">=0.9.1", optional = true}
+ipywidgets = {version = "^7.7.1", optional = true}
```
rettigl, line 42: These are required by the interactive calibration tools even without jupyter environment, so they cannot be optional.

pyproject.toml
```diff
@@ -17,10 +17,7 @@ python = ">=3.8, <3.11.9"
 bokeh = ">=2.4.2"
 dask = ">=2021.12.0"
 fastdtw = ">=0.3.4"
-fastparquet = ">=0.8.0"
```
rettigl, line 20: Why move this down? I sorted by name before...

.github/workflows/testing_multiversion.yml
```diff
@@ -26,6 +26,7 @@ jobs:
         with:
           python-version: ${{matrix.python-version}}
           poetry-version: 1.2.2
+          install-args: --all-extras
```
rettigl, line 29: Testing should be done with the minimal installation, to make sure that this works.

rettigl (changes requested):
The ipywidgets cannot be optional, because they are being used by the interactive calibration tools. Testing should also be done on the minimal requirements, to make sure this works, I would say.

rettigl:
As for the opencv question: This is a dependency via symmetrize. It was introduced here to limit the version because of a bug at some point: https://github.com/OpenCOMPES/sed/pull/111/commits/1f3b025ca47865f968eac000d4d8a83c89c0ff22
We can try removing it now again.

pyproject.toml
zain-sohail: I see. I didn't know they worked otherwise because their documentation states them working with jupyter notebooks

pyproject.toml
zain-sohail: Just a mistake

zain-sohail: But it happened because I was looking at fastparquet and already I find that including both pyarrow and fastparquet is redundant. Either one is set to optional or we remove fastparquet (https://stackoverflow.com/a/77844771)

.github/workflows/testing_multiversion.yml
zain-sohail: In future, there might be other optional dependency groups that don't impact general functionality but e.g. specific loaders. There is a possibility of including dask-jobqueue to schedule maxwell nodes through slurm for flashloader. Now this dependency wouldn't strictly be necessary for other modules. 
Or e.g. integrating multidim visualization tools such as Napari. 
In such cases, we should also allow for testing whole suite (maybe another workflow).

pyproject.toml
rettigl: Fastparquet is not explicitly mentioned, and pyarrow works similarly with dask, so I removed fastparquet now

pyproject.toml
rettigl: Not sure how much sense they make without jupyter, but the whole module cannot be loaded if they are not there, so they definitely need to be there. Maybe we can change this in the future.

.github/workflows/testing_multiversion.yml
rettigl: Then we should as you suggest provide separate tests for these, I suggest.

rettigl (approved):
LGTM. I added a small update to the release documentation.

PR #421, opened 2024-06-19
Update hextof tutorial notebook

zain-sohail:
Mostly minor changes so that users on Maxwell don't end up downloading data when it's available in the public beamtimes. The user folder is only 30 GB.

I have opted to use non-default calibration file and put that as user_config keyword. I am not certain if this is the ideal way but this can be discussed


tutorial/4_hextof_workflow.ipynb
```diff
@@ -56,10 +65,15 @@
...
    "outputs": [],
    "source": [
-    "# data_path can be defined and used to store the data in a specific location\n",
-    "dataset.get(\"Gd_W110\") # Put in Path to a storage of at least 10 Gbyte free space.\n",
-    "path = dataset.dir\n",
-    "print(path)"
+    "beamtime_dir = \"/asap3/flash/gpfs/pg2/2023/data/11019101\" # on Maxwell\n",
+    "if os.path.exists(beamtime_dir) and os.access(beamtime_dir, os.R_OK):\n",
+    "    path = beamtime_dir + \"/raw/hdf/offline/fl1user3\"\n",
+    "    buffer_path = beamtime_dir + \"/processed/tutorial/\"\n",
```
zain-sohail, line 71: if user wants to write to this folder, this won't work but since I have already populated it with buffer files, user only needs to read that data.

tutorial/4_hextof_workflow.ipynb
```diff
@@ -767,7 +739,7 @@
     "    constant=-1463.7, # this is time zero\n",
     "    flip_delay_axis=True, # invert the direction of the delay axis\n",
     "    columns=['bam'], # use the bam to offset the values\n",
-    "    weights=[0.001], # bam is in fs, delay in ps\n",
+    "    weights=[-0.001], # bam is in fs, delay in ps\n",
```
zain-sohail, line 742: We want to subtract the bam column and positive weight adds it

tutorial/4_hextof_workflow.ipynb
rettigl: This won't work then if users want to change something in the config, and thus need to recreate the buffer files. Also, does this not somewhat defy the purpose of the tutorial, if user's don't experience the full workflow (i.e. no buffer file creation)?
One could instead put just the buffer file path always into the user directly. These are fairly small for these test datasets.

tutorial/4_hextof_workflow.ipynb
rettigl: Has this been tested that it does what it's intended to do?

tutorial/4_hextof_workflow.ipynb
zain-sohail: That's true. The buffer files, being of parquet format are also quite small for the couple of runs we use in the tutorial.

tutorial/4_hextof_workflow.ipynb
zain-sohail: I have checked with Dima and this seems correct.

tutorial/4_hextof_workflow.ipynb
```diff
@@ -692,6 +663,7 @@
     "sp = SedProcessor(\n",
     "    runs=[44824,44825,44826,44827],\n",
     "    config=config_override,\n",
+    "    user_config=cal_file,\n",
```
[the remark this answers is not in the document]
zain-sohail: what if I use `folder_config=cal_file` then? Would that make sense?

tutorial/4_hextof_workflow.ipynb
```diff
@@ -84,6 +98,8 @@
    },
    "outputs": [],
    "source": [
+    "# file to store the calibration parameters in\n",
+    "cal_file = \"hextof_tutorial_cal.yaml\" \n",
```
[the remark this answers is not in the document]
zain-sohail: That is true. Nonetheless, I'd like to make it obvious that the user has the option to change the file where they load the calibration parameters from, and that it doesn't have to be sed_config.yaml. Because I had this question before.

Perhaps instead of here, we put it in the documentation then.

rettigl (approved):
LGTM

PR #422, opened 2024-06-19
add benchmarks for dataframe computation

rettigl:
Adds a benchmark for computing a one-partition dataframe for each supported reader


zain-sohail (approved):
I understand that this is to check the performance for when we update to new code? 

If we also want to compare performance between the different loaders, I guess the channel count should be same across them.

PR #429, opened 2024-06-22
using importlib to get package version

zain-sohail:
Currently after importing sed, typing `sed.__version__` gives 0.1.0 because it's been hardcoded. This PR should solve the issue.

### DIFF
--- a/.github/workflows/testing_coverage.yml
+++ b/.github/workflows/testing_coverage.yml
@@ -29,6 +29,7 @@ jobs:
       # Run pytest with coverage report, saving to xml
       - name: Run tests on python 3.8
         run: |
+          poetry install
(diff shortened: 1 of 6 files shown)


.github/workflows/testing_coverage.yml
```diff
@@ -29,6 +29,7 @@ jobs:
       # Run pytest with coverage report, saving to xml
       - name: Run tests on python 3.8
         run: |
+          poetry install
```
rettigl, line 32: Logically this would more belong to the section above, but does not matter so much.

rettigl (approved):
LGTM

.github/workflows/testing_coverage.yml
zain-sohail: I can't put a run statement with an action. But I will put it as another step.

PR #431, opened 2024-06-22
faster version of per_file channels

rettigl:
This PR implements an alternative method to generate the per_file columns, which is much more performant.
Additionally, the function to add offsets is modified for a moderate speedup (removing the map_partitions call, and rather directly apply to the df columns)


rettigl:
@zain-sohail This solution here passes the benchmarks, even with slight improvements on speed (maybe due to different worker specs), whereas the earlier code consistently is a factor 2-3 slower. I merge this here with the other branch, and then have you review it, if you want to see the changes, you can refer to this PR.

Issue #404, opened 2024-05-19
Move n_cores out of binning and into core


zain-sohail:

rettigl:
Closed by PR #451

Issue #368, opened 2024-03-20
Drop 3.8 and update to 3.9-3.12


zain-sohail:

zain-sohail:
#427 has been merged but still needs fixes so we can allow all python versions.

rettigl:
Closed by #448

Issue #173, opened 2023-10-11
tof_binning as actual binning values


steinnymir:
Fromn a discussion in #169 we decided representing tof_binning as the explicit bin size is better than using the current exponent method.
There should however be a note reporting the tof_binning value needs to be a power of 2.

current:
` tof[ns] = tof[steps] * tof_binwidth * 2**tof_binning`
Proposal:
` tof[ns] = tof[steps] * tof_binwidth * tof_binning`

zain-sohail:
Could you also put this in the documentation of binning (the binning.rst file) when a PR is created?

rettigl:
> Could you also put this in the documentation of binning (the binning.rst file) when a PR is created?

I would say it does not really belong there, at it is not related to the binning we do, but to the binning that was applied to the data by the detector readout. But it should go to the documentation, I agree. I will see where.

steinnymir:
We will discuss details about this during the Beamtime, and figure out what the correct parameters for FLASH should be.

PR #437, opened 2024-06-23
Upgrade to V1

rettigl:
Collection PR for the update to V1.
This contains breaking changes to current behavior and config file layouts.

Items to be addressed here:

- [x] Upgrade to Py3.9-3.12 #368
- [x] Energy calibration from bias values #411
- [x] implement logging #249 #490 
- [x] change binning factors from n to 2^n #173 #451  
- [x] move num_cores from binning to core #404 #451
- [x] Unify axis naming #138 
- [x] Unify config scheme (pydantic) #487
- [x] FlashLoader update/rewrite #329 
- [x] Update docs bulding scheme, and check functionality #474 
- [x] schange to `src` layout #508
- [x] Fix flash loader issues #528 
- [x] Change version to 1.0.0


rettigl:
Flash-loader is consistently slower with the new packages. Not sure why.

rettigl:
> Flash-loader is consistently slower with the new packages. Not sure why.

This directly comes from the update of dask from 2023.5.0 to 2023.12.0, tested on python 3.9.
Surprisingly, the binning in the Flash example notebook however becomes _faster_ with this new configuration (44.5s vs. 26.9 s).
As we need to look into dask anyways still, I would not consider this a problem for now.

zain-sohail:
> > Flash-loader is consistently slower with the new packages. Not sure why.
> 
> This directly comes from the update of dask from 2023.5.0 to 2023.12.0, tested on python 3.9. Surprisingly, the binning in the Flash example notebook however becomes _faster_ with this new configuration (44.5s vs. 26.9 s). As we need to look into dask anyways still, I would not consider this a problem for now.

Considering that the creation of buffer files has is independant of dask, I assume it's another package that causes this issue (maybe pandas?) 
Dask is only used when loading the buffer files and subsequent processing like sector ID seperation, forward filling, etc. and eventually binning. So it seems to be that the problem is not from dask.

zain-sohail:
Is there a reason you put 
`from __future__ import annotations`
in all files? because that was only necessary for python 3.8

rettigl:
> Dask is only used when loading the buffer files and subsequent processing like sector ID seperation, forward filling, etc. and eventually binning. So it seems to be that the problem is not from dask.

It is from dask, the performance changes if I only change the version of dask (from 2024.12.0 to 2024.5.0). There seem to be some other issues with the map_overlap stuff after loading the dataframe. see #448

.github/workflows/benchmark.yml
```diff
@@ -4,7 +4,7 @@ name: benchmark
 on:
     workflow_dispatch:
     push:
-        branches: [ main, create-pull-request/patch ]
+        branches: [ main, v1_feature_branch, create-pull-request/patch ]
```
rettigl, line 7: change before merging

.github/workflows/testing_multiversion.yml
```diff
@@ -1,9 +1,10 @@
-name: unit tests [Python 3.8|3.9|3.10|3.11]
+# Tests for all supported versions [Python 3.9|3.10|3.11|3.12]
+name: Unit Tests
 
 on:
   workflow_dispatch:
   push:
-    branches: [ main ]
+    branches: [ main, v1_feature_branch ]
```
rettigl, line 7: change before merging

rettigl:
@zain-sohail I made a pre-release for v1 from this now. If the package looks good, and the docs build also looks fine, I suggest you approve this PR, we make a final pre-v1 release and then merge and release v1, what do you think?

rettigl:
> @zain-sohail I made a pre-release for v1 from this now. If the package looks good, and the docs build also looks fine, I suggest you approve this PR, we make a final pre-v1 release and then merge and release v1, what do you think? Reviewing the whole thing again seems unfeasible to me.

@zain-sohail Did you test the package? What is your plan on how and when we can merge this?

zain-sohail:
> @zain-sohail Did you test the package? What is your plan on how and when we can merge this?

Hi Laurenz. I have tested the package in a fresh env and also looked at the docs and everything seems fine. 

Dima just wanted #534 to be merged before we put out v1. I'd hence suggest we just merge it for now. Since the lab loader doesnt have integration tests right now, I can't promise no bugs but the flash loader should run bug free. And I can fix anything later as well.

rettigl:
But what is the advantage with adding #534 still here, compared to first merging this, releasing 1.0, and then add it in 1.1? This is not a breaking change, but adds functionality. So it should be a minor release, from my point of view. This pr is already way too large, and adding here will make it more and more difficult to locate potential problems.

zain-sohail:
> But what is the advantage with adding #534 still here, compared to first merging this, releasing 1.0, and then add it in 1.1? This is not a breaking change, but adds functionality. So it should be a minor release, from my point of view. This pr is already way too large, and adding here will make it more and more difficult to locate potential problems.

That's also true but I'd actually say it is a breaking change due to me moving some hard coded values to the config files, which the users definitely need to have for the loader to work.

rettigl:
I see, I was not aware of this. But then let's complete #534 with tests (mostly dataframe module missing) and careful reviews. This might take some time though from my side, maybe @kutnyakhov also can add a review.

zain-sohail:
> I see, I was not aware of this. But then let's complete #534 with tests (mostly dataframe module missing) and careful reviews. This might take some time though from my side, maybe @kutnyakhov also can add a review.

Yes, I am also waiting on Dima's review but currently he is on holiday. I'll try to add the tests today or latest this weekend.

rettigl:
> I see, I was not aware of this. But then let's complete #534 with tests (mostly dataframe module missing) and careful reviews. This might take some time though from my side, maybe @kutnyakhov also can add a review.

Having had a brief look at the config changes in #534, they don't appear to be breaking changes to me, I. E. A config pre-534 should work with the code there (or we can make it work). Thus I would still argue to merge this one now, and then build further on that release. We should not extend this pr any further if not absolutely necessary.

zain-sohail:
> Having had a brief look at the config changes in #534, they don't appear to be breaking changes to me, I. E. A config pre-534 should work with the code there (or we can make it work). Thus I would still argue to merge this one now, and then build further on that release. We should not extend this pr any further if not absolutely necessary.

I'd suggest that we merge the config model and config file changes to v1 at least before merging with main.

rettigl:
Why? I don't see the need for this. Adding optional config elements is not a breaking change. #551 does something similar. Let's not complicate things unnecessarily.

zain-sohail:
> Why? I don't see the need for this. Adding optional config elements is not a breaking change. #551 does something similar. Let's not complicate things unnecessarily.

Because these are not optional changes to flash loaders, only for general sed they are optional. Earlier, these three fields were hard coded and I moved them out of flash.utils.get_channels and into the config.

```python
index: Optional[Union[Sequence[str], str]] = None
formats: Optional[Union[Sequence[str], str]] = None
fill_formats: Optional[Union[Sequence[str], str]] = None
```

This is definitely a breaking change in the schema at least for flash data.

rettigl:
Why don't you provide default values for these in #534 ? Then it would not break anything.

zain-sohail:
> Why don't you provide default values for these in #534 ? Then it would not break anything.

I thought we weren't including loader specific defaults in default.yaml but if that's the case then you are right.

zain-sohail (approved):

rettigl:
> > Why don't you provide default values for these in #534 ? Then it would not break anything.
> 
> I thought we weren't including loader specific defaults in default.yaml but if that's the case then you are right.

Yes, we used to handle it like this, true. I don't completely oversee why you want to make these parameters part of the config, but I suggest that you keep the code in a shape where it would also work without. That should be the case anyways for all optional parameters, I think. Maybe we should make sure that the code itself contains defaults for all optional parameters, and not the default config, which contains defaults for all required parameters. 
Anyways, we can still add this before publishing v1. I would then make a last release pre-v1, merge here, and make another release candidate for v1 from main.

.github/workflows/testing_multiversion.yml
rettigl: chaged

.github/workflows/benchmark.yml
rettigl: changed

PR #448, opened 2024-06-24
Update dask version

rettigl:
The changes to the dfops in #411 allow to pass the tests with dask2024.6.2, however there are random failures during binning (see https://github.com/dask/dask/issues/11200). This requires further inspection.

### DIFF
--- a/.github/workflows/testing_multiversion.yml
+++ b/.github/workflows/testing_multiversion.yml
@@ -13,7 +13,7 @@ jobs:
     # Using matrix strategy
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11.8", "3.12.2"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


zain-sohail:
I saw some threads on xarray about ffill algorithms and maybe they help in not using map_overlap. Putting it here because it's related
https://github.com/pydata/xarray/issues/6997
https://github.com/pydata/xarray/pull/6118

rettigl:
It depends on the dask version, not our code. Did you check the dask version you were using?

zain-sohail:
> It depends on the dask version, not our code. Did you check the dask version you were using?

Yes you are right. The 2024 version seems to cause this. Would be great to use the new optimizer they have but seems we have to figure things on our side.

rettigl:
Disabled dataframe.query-planning for now as workaround, to not have to limit the python version anymore.

zain-sohail (approved):
That seems like the best workaround for now. LGTM

PR #449, opened 2024-06-25
fix spelling in all files using Code Spell Checker

rettigl:
Fix spelling in all files


zain-sohail (approved):
LGTM

I think this can also be part of a weekly/monthly workflow like the dependency updates

rettigl:
> LGTM
> 
> I think this can also be part of a weekly/monthly workflow like the dependency updates

Not sure how this could be automatized, as you often have to do manual decisions how to correct something. But maybe there is a pre-commit hook for something like this.

PR #451, opened 2024-06-27
Tof binning as actual binning values

### LINKED ISSUE #173: tof_binning as actual binning values

### LINKED ISSUE #404: Move n_cores out of binning and into core

rettigl:
This changes the config behavior of tof_binning and adc_binning, as well as the config location of "num_cores"

closing #173 
closing #404

### DIFF
--- a/sed/binning/binning.py
+++ b/sed/binning/binning.py
@@ -267,7 +267,7 @@ def bin_dataframe(
             Defaults to None.
         pbar (bool, optional): Option to show the tqdm progress bar. Defaults to True.
         n_cores (int, optional): Number of CPU cores to use for parallelization.
-            Defaults to all but one of the available cores. Defaults to N_CPU-1.
+            Defaults to all but one of the available cores.
         threads_per_worker (int, optional): Limit the number of threads that
             multiprocessing can spawn. Defaults to 4.
         threadpool_api (str, optional): The API to use for multiprocessing.
--- a/sed/config/flash_example_config.yaml
+++ b/sed/config/flash_example_config.yaml
@@ -58,8 +58,8 @@ dataframe:
   time_stamp_alias: timeStamp
   # time length of a base time-of-flight bin in seconds
   tof_binwidth: 2.0576131995767355E-11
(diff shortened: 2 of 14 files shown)


sed/config/flash_example_config.yaml
```diff
@@ -21,8 +23,16 @@ core:
     data_parquet_dir: "tests/data/loader/flash/parquet"
 
 binning:
-  # Since this will run on maxwell most probably, we have a lot of cores at our disposal
-  num_cores: 100
+  # Histogram computation mode to use.
+  hist_mode: "numba"
```
zain-sohail, line 27: why not have these in the default config instead of each loader specific config?

sed/binning/binning.py
```diff
@@ -267,7 +267,7 @@ def bin_dataframe(
             Defaults to None.
         pbar (bool, optional): Option to show the tqdm progress bar. Defaults to True.
         n_cores (int, optional): Number of CPU cores to use for parallelization.
-            Defaults to all but one of the available cores. Defaults to N_CPU-1.
+            Defaults to all but one of the available cores.
```
zain-sohail, line 270: why not use n_cores from config in this module?

zain-sohail (approved):
LGTM

sed/binning/binning.py
rettigl: Effectively it is being used if called from the processor, but the binning functions themselves are not within classes and don't use a config.

sed/config/flash_example_config.yaml
rettigl: Effectively they are the defaults. They are just there to showcase which config parameters are available.

PR #450, opened 2024-06-27
Spell check action

rettigl:
This picks up @zain-sohail 's idea of implementing spell check pre-commit hook and github action

PR #454, opened 2024-06-29
Spelling fixes for main

rettigl:
This PR supercedes #450, and adds spell fixes, pre-commit hook and action for main. 
Might require some manual work for rebasing of the v1_feature_branch after merging.


zain-sohail (approved):
Generally fine but I noticed that the spell check marks British spelling as incorrect and prefers American. Would be good if both are accepted. But if we want to stick to American, also fine.

rettigl:
> Generally fine but I noticed that the spell check marks British spelling as incorrect and prefers American. Would be good if both are accepted. But if we want to stick to American, also fine.

I also allow British English now, but still keep the changes I made to American English for consistency.

Issue #363, opened 2024-03-18
Catch illegal kwds arguments


rettigl:
Catch and report all illegal kwds arguments in all relevant functions

zain-sohail:
How is illegal being defined?
We would nee dsome sort of schema to check that.

rettigl:
Any function that accepts **kwds should filter them depending on where they are passed to. This only becomes relevant, if you pass **kwds to a function that does not accept **kwds. Then you must ensure that only legal keyword arguments are being passed.

zain-sohail:
So there are two cases?

- **kwds are passed but shouldn't be
- **kwds are legal but some illegal ones are also passed

So I am guessing legal ones are the ones that pass the check inside the function/method?

Not sure how to approach this issue. One way would be to add decorator to each method but that doesn't seem ideal.
The other is to just perform this check at testing phase.

rettigl:
Closed by #466

PR #466, opened 2024-07-04
Catch illegal kwds

rettigl:
Addresses #363 
Would still be good to add some more tests

This is based to energy_calibration_bias_shift and includes the branch flash_tests_fix, so should be merged after these two and rebased to v1feature_branch

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -185,17 +185,17 @@ test = ["pytest (>=7.0.0)", "pytest-xdist (>=2.1.0)"]
 
 [[package]]
 name = "asteval"
-version = "0.9.33"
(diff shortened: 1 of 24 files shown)


zain-sohail (approved):
LGTM

I can't help but wonder if there's a more general way than doing a manual check everytime such as with use of decorators. But seems too tricky for current use case.

rettigl:
> I can't help but wonder if there's a more general way than doing a manual check everytime such as with use of decorators. But seems too tricky for current use case.

Maybe worth checking how this is handled in other packages, but this fixes the issue for now for us.

Issue #459, opened 2024-07-02
Refactored flashloader can't normalize data


zain-sohail:

rettigl:
This is somehow related to bam correction. In the timed_dataframe, delayStage positions are not the same as in the dataframe after bam correction, if parquets are created with the new flash loader. Currently, no idea why.

rettigl:
A major difference between the old and new loader is the dask index: It is continous in the old loader, and very sparse in the new loader:
Old: 
![grafik](https://github.com/OpenCOMPES/sed/assets/53396064/e65d1c9d-3afa-4d75-b325-3023431ac629)
![grafik](https://github.com/OpenCOMPES/sed/assets/53396064/5d50a74f-4860-44d4-acb1-948de93ed993)
New: 
![grafik](https://github.com/OpenCOMPES/sed/assets/53396064/8d7e6768-30b0-41f2-adb0-27dfd27577de)
![grafik](https://github.com/OpenCOMPES/sed/assets/53396064/f7f8c703-7e88-4caa-85a0-0203f6bd2a0c)

rettigl:
It's related to the bam correction and the mean offsetting:
Old:
![grafik](https://github.com/OpenCOMPES/sed/assets/53396064/f63f470f-f74d-4f8c-9fea-93a8d7e7e67d)
New:
![grafik](https://github.com/OpenCOMPES/sed/assets/53396064/cb17cde3-2cbc-4698-bab8-50015472d4ce)
 
Old:
![grafik](https://github.com/OpenCOMPES/sed/assets/53396064/dce1047f-dba9-44e3-95d8-9daca41379ec)
New:
![grafik](https://github.com/OpenCOMPES/sed/assets/53396064/ad2bb8f2-4eaf-43d1-9290-22c83701c033)

It seems there are a lot of empty entries in the timed_dataframe, where actually not even a photon pulse was present.

zain-sohail:
> The computation time with the parquet generated with the new loader also is >2 as long.

I think the slowdown is because of this section being offloaded to dask
https://github.com/OpenCOMPES/sed/blob/86978c08be702f550ae10c04be1357cc012ebcf0/sed/loader/flash/loader.py#L610-L614

> It seems there are a lot of empty entries in the timed_dataframe, where actually not even a photon pulse was present.

This was requested feature by Davide (#307) that the timed dataframe should contain all available values from pulses and not just ones where electrons are detected.

rettigl:
[...]

PR #469, opened 2024-07-08
Update to the BufferHandler

zain-sohail:
Fixes the issue in #459 by correctly filling the pulse resolved channels. To this end and more, this PR has the following **changes**
- Seperation of buffer files for timed and electron dataframes
  - The pulse resolved channels have the 0 values removed (TBD if it makes sense) because bam contains pulses that are invalid and stores them as 0s
  - Using concat to combine dataframe instead of join as it's faster and gives same results
  - After getting combined dataframe, df is filled on pulse and train channels
  - For electron df, all rows that don't have an electron are dropped
  - For pulse, the electron rows are dropped so we get pulse resolved data
- A class for buffer file paths which stores a list of dicts with raw, electron buffer and timed buffer filenames
- Schema check happens on both buffer types
- Changing the names from `data_raw_dir` and `data_parquet_dir` to `raw` and `processed` as these config parameters already exist under paths and hence can be used by other methods in sed, such as saving of binned data.
- Updates to metadata and tests

**Benchmarks (62 files and binning over X and Y):**
62 raw files to buffer and binning over X and Y (using 20 cores):
Old: 37.6 s ± 839 ms
New: 8.26 s ± 223 ms

62 buffer files to binning:
Old: 2.8 s ± 44.2 ms
New: 2.64 s ± 44.9 ms


rettigl:
In Tutorial 4, I get the following error when trying to load the processor:
```
	"name": "FileNotFoundError",
	"message": "No files found for run 44762 in directory ['tests/data/loader/flash']",
	"stack": "---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Cell In[15], line 1
----> 1 sp = SedProcessor(runs=[44762], config=config_override, system_config=config_file, collect_metadata=False, force_recreate=True)
      2 # You can set collect_metadata=True if the scicat_url and scicat_token are defined

File /mnt/pcshare/users/Laurenz/AreaB/sed/sed/sed/core/processor.py:164, in SedProcessor.__init__(self, metadata, config, dataframe, files, folder, runs, collect_metadata, verbose, **kwds)
    162 # Load data if provided:
    163 if dataframe is not None or files is not None or folder is not None or runs is not None:
--> 164     self.load(
    165         dataframe=dataframe,
    166         metadata=metadata,
    167         files=files,
    168         folder=folder,
    169         runs=runs,
    170         collect_metadata=collect_metadata,
    171         **kwds,
    172     )

File /mnt/pcshare/users/Laurenz/AreaB/sed/sed/sed/core/processor.py:410, in SedProcessor.load(self, dataframe, metadata, files, folder, runs, collect_metadata, **kwds)
    402         dataframe, timed_dataframe, metadata = self.loader.read_dataframe(
    403             folders=cast(str, self.cpy(folder)),
    404             runs=runs,
   (...)
    407             **kwds,
    408         )
    409     else:
--> 410         dataframe, timed_dataframe, metadata = self.loader.read_dataframe(
    411             runs=runs,
    412             metadata=metadata,
    413             collect_metadata=collect_metadata,
    414             **kwds,
    415         )
    417 elif folder is not None:
    418     dataframe, timed_dataframe, metadata = self.loader.read_dataframe(
    419         folders=cast(str, self.cpy(folder)),
    420         metadata=metadata,
    421         collect_metadata=collect_metadata,
    422         **kwds,
    423     )

File /mnt/pcshare/users/Laurenz/AreaB/sed/sed/sed/loader/flash/loader.py:303, in FlashLoader.read_dataframe(self, files, folders, runs, ftype, metadata, collect_metadata, detector, force_recreate, processed_dir, debug, **kwds)
    301 runs_ = [str(runs)] if isinstance(runs, (str, int)) else list(map(str, runs))
    302 for run in runs_:
--> 303     run_files = self.get_files_from_run_id(
    304         run_id=run,
    305         folders=self.raw_dir,
    306     )
    307     files.extend(run_files)
    308 self.runs = runs_

File /mnt/pcshare/users/Laurenz/AreaB/sed/sed/sed/loader/flash/loader.py:170, in FlashLoader.get_files_from_run_id(self, run_id, folders, extension)
    168 # Check if any files are found
    169 if not files:
--> 170     raise FileNotFoundError(
    171         f\"No files found for run {run_id} in directory {str(folders)}\",
    172     )
    174 # Return the list of found files
    175 return [str(file.resolve()) for file in files]

FileNotFoundError: No files found for run 44762 in directory ['tests/data/loader/flash']"
```

benchmarks/benchmark_sed.py
```diff
@@ -188,6 +188,7 @@ def test_loader_compute(loader: BaseLoader) -> None:
         loaded_dataframe, _, loaded_metadata = loader.read_dataframe(
             runs=runs[loader_name],
             collect_metadata=False,
+            force_recreate=True,
```
rettigl, line 191: This should not work anymore for all other loaders when #466 is merged.

sed/config/flash_example_config.yaml
```diff
@@ -18,9 +18,9 @@ core:
   # provided, the loader will try to find the data based on year beamtimeID etc
   paths:
     # location of the raw data.
-    data_raw_dir: "tests/data/loader/flash/"
+    raw: "tests/data/loader/flash/"
     # location of the intermediate parquet files.
-    data_parquet_dir: "tests/data/loader/flash/parquet"
+    processed: "tests/data/loader/flash/parquet"
```
rettigl, line 23: Misses update of Tutorial

sed/loader/flash/buffer_handler.py
```diff
@@ -12,11 +11,70 @@
...
+    """
+    A class for handling the paths to the raw and buffer files of electron and timed dataframes.
+    A list of file sets (dict) are created for each H5 file containing the paths to the raw file
+    and the electron and timed buffer files.
+
+    Structure of the file sets:
+    {
+        "raw": Path to the H5 file,
+        "electron": Path to the electron buffer file,
+        "timed": Path to the timed buffer file,
```
rettigl, line 32: Is it really necessary to store the data twice (actually three times)? This appears very inefficient to me. Why is this necessary?

sed/loader/flash/buffer_handler.py
```diff
@@ -79,160 +129,137 @@ def _schema_check(self) -> None:
...
 
-        # Get the list of H5 files to read and the corresponding buffer files to create
-        self.missing_h5_files = list(compress(h5_paths, files_to_read))
-        self.save_paths = list(compress(self.buffer_paths, files_to_read))
+        # Create a DataFrameCreator instance and the h5 file
+        df = DataFrameCreator(config_dataframe=self._config, h5_path=paths["raw"]).df
 
-        print(f"Reading files: {len(self.missing_h5_files)} new files of {len(h5_paths)} total.")
+        # forward fill all the non-electron channels
+        df[self.fill_channels] = df[self.fill_channels].ffill()
```
rettigl, line 148: Do we need to fill them here already before saving/loading? This used to be done by ffill_lazy, no?

sed/loader/flash/buffer_handler.py
```diff
@@ -79,160 +129,137 @@ def _schema_check(self) -> None:
...
+        df.dropna(subset=electron_channels).astype(dtypes).reset_index().to_parquet(
+            paths["electron"],
+        )
 
-        # Reset the index of the DataFrame and save it as a parquet file
-        df.reset_index().to_parquet(parquet_path)
+        # timed dataframe
+        dtypes = get_dtypes(self._config["channels"], ["per_pulse", "per_train"])
+        # drop the electron channels and only take rows with the first electronId
+        df[self.fill_channels].loc[:, :, 0].astype(dtypes).reset_index().to_parquet(paths["timed"])
```
rettigl, line 161: I don't understand why all these operations cannot be done lazily on the loaded dataframe from one source.

sed/loader/flash/buffer_handler.py
```diff
@@ -79,160 +129,137 @@ def _schema_check(self) -> None:
...
-                df_electron,
-                config=self._config,
+            # Forward fill the non-electron channels across files
+            overlap = min(file["num_rows"] for file in file_stats[typ].values())
+            iterations = self._config.get("forward_fill_iterations", 2)
+            df = forward_fill_lazy(
+                df=df,
+                columns=self.fill_channels,
+                before=overlap,
+                iterations=iterations,
```
rettigl, line 206: You still do this here again?

sed/loader/flash/buffer_handler.py
```diff
@@ -79,160 +129,137 @@ def _schema_check(self) -> None:
...
+                    extend_aux=True,
+                ),
+            ) - {"electronId"}
+            self._schema_check(self.fp["timed"], schema_set)
+
+        self._save_buffer_files(force_recreate, debug)
 
         self._fill_dataframes()
+
+        # Correct the 3-bit shift which encodes the detector ID in the 8s time
```
rettigl, line 259: your naming is somewhat misleading here. You save the buffer files, but you never load them...

rettigl (changes requested):
I tested this out now. The electron df seems to work correctly, but the computation of the normalization histograms takes a lot longer now (~33s vs. +24s on main). Also, the normalization does not look as good as before:
main:
grafik
fix-459:
grafik

It could off course be that the earlier normalization effectively only normalizes by total number of electrons, and not time...

benchmarks/benchmark_sed.py
zain-sohail: Yes, but this is a keyword specific to Flash so I don't think it'll be passed when using other loaders. Or do you mean the check happens earlier?

sed/config/flash_example_config.yaml
zain-sohail: Thanks. Also needs updating with the main branch

sed/loader/flash/buffer_handler.py
zain-sohail: I'd consider the buffer files the raw files for our case as the flash raw h5 files need to be restructured and filled. 

Of course seperately storing electron and timed data seems redundant. But like we saw from your benchmarks, dask seems to perform rather poorly when the indexes are not monotonic. And they aren't if we drop nans after loading the data. 

Moreover, this now gives us correct counts from the parquet metadata. And that data, we'd definitely like to use later.

sed/loader/flash/buffer_handler.py
zain-sohail: Right. Can be fixed.

sed/loader/flash/buffer_handler.py
zain-sohail: This was for filling between different files. The previous operation is filling within files. 

I will take a look again.

benchmarks/benchmark_sed.py
rettigl: Like you defined it, it will be passed to all readers. Right now it is being ignored, but will raise after merging #466

sed/loader/flash/buffer_handler.py
zain-sohail: I have tested it again lazily. The lazy operations take about 4x longer compared to this version.
Dask Lazy:
image
files are saved to parquet after resetting index straight to parquet and then loaded by dask. The timing is after that.

This PR:
image

A single dataframe I store with all nans also has a filesize of about 5 MB. The electron and timed parquets size about 2.5 MB each, which is the same storage size.

I used run 44824 which has 20 files.

sed/loader/flash/buffer_handler.py
zain-sohail: Yes but then I can't drop nans in subsequent steps because filling is necessary to have correct information before dropping.
And as mentioned in other comment, dropping nans operation is really slow in dask. Likely because it modifies the df length which destroys the indexing (indexed operations are always faster).

zain-sohail:
From my testing, the main reason dask performs so poorly has to do with index sorting (This is also why the data is never having divisions in dask if you notice). Seems like flash daq just writes when possible without having a consistent timing in files. Seems to specially occur with per_train channels. So even though I was sorting individual files in the Dataframe class, another sort needs to happen across an entire run I believe. 

Here you can see two different files. The first one contains trains 74 and 76. and then next one starts at 75. So it becomes unsorted when these files are merged

Screenshot 2024-07-16 at 22 40 28

Between runs shouldn't be an issue since there is a gap which should be sufficient to not have this weird overlap. However, even sorting the run seems extremely slow, if done with dask, because the sort operation requires extensive shuffling.
Alternate is to do it in memory, which should be possible on maxwell but definitely not locally.

sed/loader/flash/buffer_handler.py
rettigl: The difference between the "old" version, and here is 209 vs. 366 mbyte, while the current version in v1 branch is 500 mbyte. So indeed, this is actually an improvement. Slightly larger compared to the main version, but acceptable I would say (it also contains more information, i.e. the shots without electrons)

sed/loader/flash/buffer_handler.py
rettigl: Then let's keep this approach (although I don't completely understand)

sed/loader/flash/buffer_handler.py
rettigl: You are right.

sed/loader/flash/buffer_handler.py
rettigl: fine.

sed/config/flash_example_config.yaml
rettigl: done that.

rettigl:
> Edited: Index sorting does not improve speed with using dask. I find that the current approach is the fastest.

WIth respect to what you posted earlier here: I think it does not matter if the sub-indices are unsorted. For dask, there is only one main index which should ideally be gapless and sorted, no?

rettigl:
Histogram of timeStamp (should be flat)
Old:
grafik
New:
grafik

rettigl (approved):
LGTM now.

rettigl:
> > WIth respect to what you posted earlier here: I think it does not matter if the sub-indices are unsorted. For dask, there is only one main index which should ideally be gapless and sorted, no?
> 
> I was thinking setting the index to trainId would help, since otherwise, there is no monotonic index for dask to work with (it is only monotonic within a file and then restart). But weirdly, the operations take longer with index. I find that rather unexpected behavior.

I think the natural dask index is per partition, as every partition is like a separate pandas df. How would you use trainId as a unique index? Anyways, what would you need a continous index across partitions for?

zain-sohail:
> I think the natural dask index is per partition, as every partition is like a separate pandas df. How would you use trainId as a unique index? Anyways, what would you need a continous index across partitions for?

Yes that's true but if the index exists and is sorted, dask can calculate divisions. It is supposed to avoid expensive data shuffling (See https://docs.dask.org/en/stable/dataframe-best-practices.html#use-the-index)
But I believe since we mostly work with shared memory (not entirely sure), it might not make an impact.

rettigl:
@zain-sohail I think this PR and #459 contain the relevant information for reverting back to the old behavior

PR #471, opened 2024-07-10
Hextof tutorial missed out changes

zain-sohail:
Accidently reverted some changes that needed to be there.

Moreover, using the tutorial beamtime_id now because before it was set to the test data beamtime.

Additionally, the maxwell paths are provided instead of test data paths.

### DIFF
--- a/sed/config/flash_example_config.yaml
+++ b/sed/config/flash_example_config.yaml
@@ -6,7 +6,7 @@ core:
   # the beamline where experiment took place
   beamline: pg2
   # the ID number of the beamtime
-  beamtime_id: 11013410
+  beamtime_id: 11019101
   # the year of the beamtime
   year: 2023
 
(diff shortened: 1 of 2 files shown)


sed/config/flash_example_config.yaml
```diff
@@ -5,18 +5,18 @@ core:
...
 
   # The paths to the raw and parquet data directories. If these are not
   # provided, the loader will try to find the data based on year beamtimeID etc
   paths:
     # location of the raw data.
-    data_raw_dir: "tests/data/loader/flash/"
+    data_raw_dir: "/asap3/flash/gpfs/pg2/2023/data/11019101/raw/hdf/offline/fl1user3"
     # location of the intermediate parquet files.
-    data_parquet_dir: "tests/data/loader/flash/parquet"
+    data_parquet_dir: "Gd_W110/processed/"
```
rettigl, line 19: I would suggest to leave them empty as we are anyways changing them

sed/config/flash_example_config.yaml
```diff
@@ -123,8 +123,8 @@ dataframe:
       group_name: "/uncategorised/FLASH.EXP/HEXTOF.DAQ/DLD1/"
       slice: 3
 
-    # The auxillary channel has a special structure where the group further contains
-    # a multidim structure so further aliases are defined below
+    # The auxiliary channel has a special structure where the group further contains
+    # a multidimensional structure so further aliases are defined below
```
rettigl, line 127: These should vanish if you propperly rebase

rettigl (approved):
Generally LGTM, but needs to be rebased. And I suggest to remove the paths in the config/set to empty strings.

PR #474, opened 2024-07-16
Docs website change: Pydata docs theme

zain-sohail:
This PR changes the current readthedocs theme we are using for docs to the pydata docs theme, which supports a lot more features.
Here, we also change the documentation github actions workflow as after building, it will push that versions (like 0.2 etc.) to the docs repo in the sed folder. 

The latest main version can be found here: https://opencompes.github.io/docs/sed/latest/

There are some changes to the structure of the documentation (such as seperating user guide, advanced workflows and api) that I saw made more sense but these can be discussed here.

Current main branch is supporting python 3.8 but the pydata theme only supports >3.9 so the workaround is to pip install on the CI for now but in v1, we can change this.


.cspell/custom-dictionary.txt
```diff
@@ -403,3 +403,7 @@ pypi
 pypa
 cobertura
 coverallsapp
+viewcode
+pydata
+ALLUSERPROFILE
```
rettigl, line 408: This is a typo, must be ALLUSERSPROFILE

.github/workflows/documentation.yml
```diff
@@ -84,25 +66,63 @@ jobs:
...
+          path: sphinx-docs
+
+      - name: Determine version folder
+        id: version-folder
+        run: |
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/}
+            echo "folder=sed/$VERSION" >> $GITHUB_OUTPUT
+          else
+            echo "folder=sed/latest" >> $GITHUB_OUTPUT
```
rettigl, line 113: This will alyways set any manual run of this workflow as the latest version. I suggest to add a case for generating docs for specific branches for testing.

.github/workflows/documentation.yml
```diff
@@ -84,25 +66,63 @@ jobs:
...
+          fi
+
+      - name: Copy documentation to the right version folder
+        run: |
+          mkdir -p docs-repo/${{ steps.version-folder.outputs.folder }}
+          cp -r sphinx-docs/* docs-repo/${{ steps.version-folder.outputs.folder }}
```
rettigl, line 119: This contains a substantial amount of unnecceary data (>50 mbyte per generated docs version). One could e.g. omitt the .doctrees folder, and the ipynb files from the tutorials folder.

docs/conf.py
```diff
@@ -98,7 +101,28 @@ def _get_project_meta():
...
+    "navbar_center": ["version-switcher", "navbar-nav"],
+    "show_nav_level": 2,
+    "show_version_warning_banner": True,
+    # maybe better to use _static/switcher.json on github pages link instead of the following
+    "switcher": {
+        "json_url": "https://raw.githubusercontent.com/OpenCOMPES/docs/main/sed/switcher.json",
+        "version_match": version,
+    },
+    "content_footer_items": ["last-updated"],
+}
```
rettigl, line 118: I don't like the curent layout so much, which does not show the whole docs tree in the left sidebar (e.g., if you are in tutorials, you don't see the main menu entries). can this be changed to look similar to the old theme?

docs/user_guide/config.md
```diff
@@ -0,0 +1,29 @@
...
+It will load an (optional) provided config file, or alternatively use a passed python dictionary as initial config dictionary, and subsequently look for the following additional config files to load:
+
+* ``folder_config``: A config file of name :file:`sed_config.yaml` in the current working directory. This is mostly intended to pass calibration parameters of the workflow between different notebook instances.
+* ``user_config``: A config file provided by the user, stored as :file:`.sed/config.yaml` in the current user's home directly. This is intended to give a user the option for individual configuration modifications of system settings.
+* ``system_config``: A config file provided by the system administrator, stored as :file:`/etc/sed/config.yaml` on Linux-based systems, and :file:`%ALLUSERPROFILE%/sed/config.yaml` on Windows. This should provide all necessary default parameters for using the sed processor with a given setup. For an example for an mpes setup, see :ref:`example_config`
```
rettigl, line 8: ALLUSERSPROFILE
Why did you introduce this typo here again?

docs/user_guide/index.md
```diff
@@ -0,0 +1,25 @@
...
+:caption: Installing SED
+:maxdepth: 1
+installation
+```
+
+```{toctree}
+:caption: Basic concepts
+../tutorial/1_binning_fake_data
+../tutorial/8_jittering_tutorial
+../tutorial/3_metadata_collection_and_export_to_NeXus
```
rettigl, line 19: I don't find these helpful here in this order. They hardly can be understood here without the other tutorials.

docs/user_guide/index.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+```{toctree}
+:caption: Advance
+config
```
rettigl, line 24: Why is config advanced? This should be in the basic concepts

rettigl: Explanation for Jittering is maybe more something advanced

docs/user_guide/installation.md
```diff
@@ -0,0 +1,74 @@
...
+.sed-dev\Scripts\activate
+```
+
+3. Install the repository in editable mode with all dependencies:
+
+```bash
+pip install -e .[all]
+```
+
+Now you have the development version of `sed` installed in your local environment. Feel free to make changes and submit pull requests.
```
rettigl, line 74: Why did you remove poetry docs?

docs/workflows/index.md
```diff
@@ -0,0 +1,15 @@
...
+      Workflows showcasing different SED loaders and methods
+---
+# Workflows
+
+```{toctree}
+:caption: Basic concepts
+../tutorial/2_conversion_pipeline_for_example_time-resolved_ARPES_data
+../tutorial/4_hextof_workflow
+../tutorial/6_binning_with_time-stamped_data
+../tutorial/7_correcting_orthorhombic_symmetry
```
rettigl, line 14: This separation does not really make sense to me.

rettigl (changes requested):
I generally like the idea behind the new workflow, but suggest some improvements to the structure and layout. See detailed comments.

docs/user_guide/config.md
zain-sohail: I think I included the typo in cspell dict and it propogated it here

docs/conf.py
zain-sohail: It should be possible to change that, I believe. However, those entries are showing on the right side, which I find quite a cleaner layout. And I can directly navigate to other tutorials without issue.

docs/user_guide/index.md
zain-sohail: That's true. Was just an initial idea. I think this explanation with each instrument is quite a lot and unnecessary for a new user. But we should definitely include a section to describe creating a new configuration.

And regarding jittering, I thought it's part of the standard workflow so would be good to be in basics, perhaps also a basic intro.

docs/user_guide/installation.md
zain-sohail: I moved that to the maintainer section 
https://opencompes.github.io/docs/sed/latest/misc/maintain.html

I can of course put it back but it seems most users don't want to use poetry. Perhaps linking that page in this file makes sense?

docs/workflows/index.md
zain-sohail: I think it makes sense to have an introduction to the different aspects of the library, irrespective of the different loaders/workflows at different instruments.
What is your suggestion?

docs/user_guide/index.md
zain-sohail: I get the point. But I think we need to include more documentation about the library and the workflow tutorials are for specific cases.
For example, I wouldn't use the WSe2 entire dataset for jittering example but we can take a small subset of it and show how jittering works similar to tutorial 1.

docs/conf.py
zain-sohail: See right side
image

rettigl: What I mean is that these menues are all distributed over the page, i.e. main menu on the top, section menue on the left, page menu on the right. I find this very confusing, and liked the unique tree of the old docs much better

docs/user_guide/installation.md
rettigl: That's probably fine.

docs/conf.py
zain-sohail: Yes, but this seems to be how basically all modern projects in python are documented. See [numpy](https://numpy.org/doc/stable/index.html), [scipy](https://docs.scipy.org/doc/scipy/), [xarray](https://docs.xarray.dev/en/stable/), [dask](https://docs.dask.org/en/stable/). I think dask keeps the page menu on both left and right siden and some others might too. 
Since our library is based on these packages, I think it's more intuitive for user to follow such layout. We can of course change it but including complete heirarchy in the left menu clutters everything, I find.

rettigl: I can't follow. In the examples you provide, the left menu always stays the same, independent where you navigate. For your new layout, this is not the case. This is very confusing, and makes it difficult to get an overview of the docs. This is my point.

.github/workflows/documentation.yml
rettigl: added lines to remove this code

docs/conf.py
rettigl: We can probably keep the current layout, changing that appears to be not so straight-forward

docs/workflows/index.md
rettigl: I reordered things slightly, and added a section advanced topics

docs/user_guide/index.md
rettigl: Restructured

rettigl (approved):
Fnished from my side

zain-sohail (commented):
I have taken a look at the develop docs and looks good. Any doc changes, we can always do later. I'd say we merge and see if the switching is working properly. Otherwise we might need to delete the static files and rebuild them.

PR #479, opened 2024-07-18
Flash minor changes (Merge to #469)

zain-sohail:
Some minor refactoring. Should be merged to fix-459 and then to V1. Mainly:

- Allowing for a auxiliary channel alias and it's subchannels. The structure is generally not my favorite but I can't think of a better way as of now. Suggestions welcome.
- All col metadata is fetched from parquets now. Will be useful to display info later.
- Minor structural changes to how this column metadata is represented
- Some naming changes in the buffer module
- Buffer module now uses dtypes to save to correct dtype (maybe nans in timed df need be dropped to get correct dtypes)


sed/loader/flash/dataframe.py
```diff
@@ -257,10 +259,18 @@ def df_train(self) -> pd.DataFrame:
             # contains multiple channels inside. Even though they are resolved per train,
             # they come in pulse format, so the extra values are sliced and individual channels are
             # created and appended to the list
+            aux_alias = self._config.get("aux_alias", "dldAux")
+            sub_channels = self._config.get("aux_subchannels_alias", "dldAuxChannels")
             if channel == "dldAux":
```
rettigl, line 264:
```suggestion
            if channel == aux_alias:
```

sed/config/flash_example_config.yaml
```diff
@@ -150,13 +160,27 @@ dataframe:
...
-        sampleBias: 0
-        tofVoltage: 1
-        extractorVoltage: 2
-        extractorCurrent: 3
-        cryoTemperature: 4
-        sampleTemperature: 5
-        dldTimeBinSize: 15
+        sampleBias:
+          slice: 0
+          dtype: float32
```
rettigl, line 165: If you treat them that way with "slice", can't you just tread them as individual channels and drop the extra handling altogether?

sed/loader/flash/buffer_handler.py
```diff
@@ -149,16 +146,17 @@ def _save_buffer_file(self, paths: dict[str, Path]) -> None:
 
         # Reset the index of the DataFrame and save both the electron and timed dataframes
         # electron resolved dataframe
-        electron_channels = get_channels(self._config["channels"], "per_electron")
-        dtypes = get_dtypes(self._config["channels"], "all")
+        electron_channels = get_channels(self._config, "per_electron")
+        dtypes = get_dtypes(self._config["channels"], df.columns.values)
```
rettigl, line 150: It would be consistent to also change the first argument of get_dtypes to accept the whole config.

sed/loader/flash/dataframe.py
```diff
@@ -257,10 +259,18 @@ def df_train(self) -> pd.DataFrame:
...
-                aux_channels = self._config["channels"]["dldAux"]["dldAuxChannels"].items()
-                for name, slice_aux in aux_channels:
-                    series.append(pd.Series(dataset[: key.size, slice_aux], index, name=name))
+                aux_channels = self._config["channels"][aux_alias][sub_channels].items()
+                for name, values in aux_channels:
```
rettigl, line 266: Why define aux_channels here? Can't you just directly iterate the dict?

sed/loader/flash/loader.py
```diff
@@ -220,15 +220,16 @@ def get_elapsed_time(self, fids: Sequence[int] = None, **kwds) -> float | list[f
             raise KeyError(
                 "File statistics missing. Use 'read_dataframe' first.",
             ) from exc
+        time_stamp_alias = self._config["dataframe"].get("time_stamp_alias", "timeStamp")
```
rettigl, line 223: `We should add this to the default config. It is also required e.g. in the normalization histogram function for timestamps

sed/loader/flash/utils.py
```diff
@@ -68,35 +71,39 @@ def get_channels(
...
                     )
                 else:
-                    channels.extend([DLD_AUX_ALIAS])
+                    channels.extend([dld_aux_alias])
 
     return channels
 
 
-def get_dtypes(channels_dict: dict, formats: str | list[str]) -> dict:
+def get_dtypes(channels_dict: dict, df_cols: list) -> dict:
```
rettigl, line 89: see above, pass dataframe_config

sed/loader/flash/utils.py
```diff
@@ -68,35 +71,39 @@ def get_channels(
...
-    }
+    dtypes = {}
+    for channel in df_cols:
+        try:
+            dtypes[channel] = channels_dict[channel].get("dtype")
+        except KeyError:
+            try:
+                dtypes[channel] = channels_dict["dldAux"][channel].get("dtype")
+            except KeyError:
+                dtypes[channel] = None
```
rettigl, line 108: Put as default float32 here?

sed/loader/flash/buffer_handler.py
```diff
@@ -214,6 +217,13 @@ def _fill_dataframes(self):
...
         self.metadata.update({"file_statistics": file_stats, "filling": filling})
+        # Correct the 3-bit shift which encodes the detector ID in the 8s time
+        if self._config.get("split_sector_id_from_dld_time", False):
+            self.df["electron"], meta = split_dld_time_from_sector_id(
+                self.df["electron"],
+                config=self._config,
+            )
+            self.metadata.update(meta)
 
     def run(
```
rettigl, line 228: I'd suggest to make run return the two dataframes directly, then there is no need to access the class attributes. Maybe also rename to e.g. process_and_load_dataframe or whatever.

rettigl (approved):
Looks all good to me, a few suggestions in the comments.

sed/config/flash_example_config.yaml
zain-sohail: The issue is we already need to slice the dld channels first (See line 161). So it's a nested array. 
Probably np.take can take a list of slice indexes. I can take a look at that.

sed/loader/flash/dataframe.py
zain-sohail: Yes surely. Just makes the argument really long.

sed/config/flash_example_config.yaml
zain-sohail: I looked into doing it another way but it makes it way more complex. So for now, I am just keeping this but generalized it to be called subChannels

sed/loader/flash/utils.py
zain-sohail: I think it's a bit risky if the data is coerced to 32 without users knowledge. I am not sure if a warning is provided by pandas if the data is float64.

PR #490, opened 2024-08-13
Logging

rettigl:
This PR converts all output into logger entries.
It also adds a call logger to record all function calls on the main workflow functions in the processor class

### DIFF
--- a/sed/calibrator/delay.py
+++ b/sed/calibrator/delay.py
@@ -48,8 +48,8 @@ class DelayCalibrator:
         else:
             self._config = {}
 
-        self.verbose = verbose
-        set_verbosity(logger, self.verbose)
+        self._verbose = verbose
+        set_verbosity(logger, self._verbose)
 
         self.adc_column: str = self._config["dataframe"].get("adc_column", None)
         self.delay_column: str = self._config["dataframe"]["delay_column"]
@@ -59,6 +59,25 @@ class DelayCalibrator:
         )
         self.calibration: dict[str, Any] = self._config["delay"].get("calibration", {})
         self.offsets: dict[str, Any] = self._config["delay"].get("offsets", {})
+
+    @property
+    def verbose(self) -> bool:
+        """Accessor to the verbosity flag.
+
+        Returns:
+            bool: Verbosity flag.
+        """
+        return self._verbose
+
+    @verbose.setter
+    def verbose(self, verbose: bool):
+        """Setter for the verbosity.
+
+        Args:
+            verbose (bool): Option to turn on verbose output. Sets loglevel to INFO.
+        """
+        self._verbose = verbose
+        set_verbosity(logger, self._verbose)
 
     def append_delay_axis(
         self,
--- a/sed/calibrator/delay.py
+++ b/sed/calibrator/delay.py
@@ -326,15 +326,13 @@ class DelayCalibrator:
             weights = []
             preserve_mean = []
             reductions = []
-            if not suppress_output:
-                logger.info("Delay offset parameters:")
+            log_str = "Delay offset parameters:"
             for k, v in offsets.items():
                 if k == "creation_date":
                     continue
                 if k == "constant":
                     constant = v
-                    if not suppress_output:
-                        print(f"   Constant: {constant} ")
+                    log_str += f"\n   Constant: {constant}"
                 elif k == "flip_delay_axis":
                     fda = str(v)
                     if fda.lower() in ["true", "1"]:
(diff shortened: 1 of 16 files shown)


sed/core/logging.py
```diff
@@ -19,26 +21,55 @@
...
+    set_base_handler: bool = False,
+    user_log_path: str | None = None,
 ) -> logging.Logger:
     """
     Configures and returns a logger with specified log levels for console and file handlers.
 
     Args:
         name (str): The name of the logger.
-        user_log_path (str): Path to the user-specific log directory. Defaults to DEFAULT_LOG_DIR.
+        base_name (str, optional): The name of the base logger. Defaults to "sed".
```
zain-sohail, line 32: This arg is not present. I believe you mean set_base_handler

sed/core/logging.py
```diff
@@ -51,24 +82,26 @@ def setup_logging(
...
 
-    # Create formatter for file
-    file_formatter = logging.Formatter(
-        "%(asctime)s - %(name)s - %(levelname)s - %(message)s in %(filename)s:%(lineno)d",
-    )
-    file_handler.setFormatter(file_formatter)
 
-    # Add file handler to logger
-    logger.addHandler(file_handler)
+def call_logger(logger: logging.Logger):
```
zain-sohail, line 91: I like the idea of logging calls. As debug it also seems good. But we can additionally also add this info to the metadata (only the function call and parameters, excluding other details)

sed/core/logging.py
```diff
@@ -19,26 +21,55 @@
 
 def setup_logging(
```
zain-sohail, line 22: I generally think we need to set the default logging to INFO and allow for an easy way to change that to debug. Maybe something like this way, the user can change to their preference. 
```
from sed.logging import VERBOSITY
VERBOSITY = 10 # to set to debug
``` 

Maybe doesn't make sense to log inside log files, unless specified by user.

sed/calibrator/delay.py
```diff
@@ -22,22 +28,33 @@ class DelayCalibrator:
...
+                Defaults to True.
         """
         if config is not None:
             self._config = config
         else:
             self._config = {}
 
+        self.verbose = verbose
+        if self.verbose:
+            logger.handlers[0].setLevel(INFO)
```
zain-sohail, line 54: It is not quite clear by [0] should be used.

sed/calibrator/delay.py
```diff
@@ -92,8 +109,7 @@ def append_delay_axis(
                 Defaults to config["delay"]["p2_key"]
             t0_key (str, optional): hdf5 key for t0 value (mm).
                 Defaults to config["delay"]["t0_key"]
-            verbose (bool, optional): Option to print out diagnostic information.
-                Defaults to True.
+            suppress_output (bool, optional): Option to suppress log output. Defaults to False.
```
zain-sohail, line 112: Why is suppress output flag needed? If it's set to True, not even at debug level, this information will be visible.
I'd say we can set all print to logger.info like you did without the `not suppress_output` conditional. Then, if user selects verbose, it prints out the information (already done). And for debug, it uses what I wrote in another comment.

sed/calibrator/delay.py
```diff
@@ -13,6 +15,10 @@
 import pandas as pd
 
 from sed.core import dfops
+from sed.core.logging import setup_logging
+
+# Configure logging
+logger = setup_logging("delay")
```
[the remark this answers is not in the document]
rettigl: I set up a hierarchy of loggers, i.e. the base logger writes to the file, and the child loggers print to the output. With the module name, this hierarchy gets more complicated, and the output longer. But could probably also work, if you prefer.

sed/core/logging.py
rettigl: True, needs to be updated

sed/core/logging.py
rettigl: Yes, this can be something for the future, to build up a call stack to reproduce a pipeline, similar to what @steinnymir had in mind.

sed/core/logging.py
rettigl: I did not like the idea to expose the loglevels directly to the user, thus the verbosity flag. There should still be a setter for it that updates the loglevel globally.

sed/calibrator/delay.py
rettigl: There is only one handler in each logger, which is always [0].

sed/calibrator/delay.py
rettigl: This is used for the timed dataframes. If they are present, each output appears twice otherwise.

zain-sohail (approved):
Good to merge. Perhaps the comments I mentioned can be looked at, or can be done at another time.

sed/core/logging.py
rettigl: done

sed/core/logging.py
rettigl: In favor of not enlarging this PR, I create an issue for this.

sed/core/logging.py
rettigl: I added a setter for the verbosity level now.

PR #516, opened 2024-11-12
Update jupyter

rettigl:
Update jupyterlab and corresponding widgets to work with jupyterlab >4 (has been updated on maxwell).


rettigl:
@zain-sohail The jupyter lab version on Maxwell has been updated to 4.1.x in the py3.11 environment on maxwell. This upgrades the widget packages to be compatible to that.

zain-sohail:
> Update jupyterlab and corresponding widgets to work with jupyterlab >4 (has been updated on maxwell).

The widgets seem to work for me in the current v1 branch. Could you elaborate the updates?

rettigl:
> > Update jupyterlab and corresponding widgets to work with jupyterlab >4 (has been updated on maxwell).
> 
> The widgets seem to work for me in the current v1 branch. Could you elaborate the updates?

For Testing, make sure you are using the kernel from the right environment (py311 vs. py39)
This is what I get with the package config of v1, if I start the Python 3.11 setting:
grafik
grafik
grafik
After installing the package config from this branch, it works:
grafik

Similarly, if I start with python 3.9 setting:
grafik
grafik
It does not work with this branch:
grafik
But does with the v1 branch:
grafik

zain-sohail:
> For Testing, make sure you are using the kernel from the right environment (py311 vs. py39) This is what I get with the package config of v1, if I start the Python 3.11 setting:

So indeed I loaded this env but for creating a poetry or pip env, did you use python 3.9 or 3.11? Because currently I load jupyterlab 4 with this python 3.11 option at start screen (but using the older python 3.9 for my sed env), everything works from v1:

image
image

rettigl:
> > For Testing, make sure you are using the kernel from the right environment (py311 vs. py39) This is what I get with the package config of v1, if I start the Python 3.11 setting:
> 
> So indeed I loaded this env but for creating a poetry or pip env, did you use python 3.9 or 3.11? Because currently I load jupyterlab 4 with this python 3.11 option at start screen (but using the older python 3.9 for my sed env), everything works from v1:
> 
> image image

I used the python version that is active per default with the setting. The problem is not with plotting, this works in both cases, it is about the widgets, I. E. The progress bars or control inputs for the Gui panels

zain-sohail:
> I used the python version that is active per default with the setting. The problem is not with plotting, this works in both cases, it is about the widgets, I. E. The progress bars or control inputs for the Gui panels

Widgets such as progress bars also work fine. 
image
I also tested this with a new envusing python 3.11 and the widgets work fine. 

We can still update this if it's causing problems on your end. Does this branch also work with jupyter <4 (or the python 3.9 jupyter that maxwell offers)?

rettigl:
> I also tested this with a new envusing python 3.11 and the widgets work fine.

I find this different behavior quite strange, but if it also works for you with the upgraded packages, I suggest to merge this.

> Does this branch also work with jupyter <4 (or the python 3.9 jupyter that maxwell offers)?

See my examples above. For me, this branch works for jupyterlab > 4, while the v1 branch does not, and vice versa.

As the only reason I limited jupyterlab was the installation on maxwell, and there is now a newer version there, I don't see a reason anymore to limit it.

zain-sohail (approved):

Issue #270, opened 2023-11-14
fit_spot_size: visual tool to track optical and FEL beam overlaps and obtain effective fluence.


steinnymir:

zain-sohail:
#273 #272 merge these PRs

PR #518, opened 2024-11-18
New tutorials for hextof

### LINKED ISSUE #426: Energy calibration for core levels using side bands
It would be nice to add the possibility of using side bands measured at t0 for energy calibration.
If we e.g. bin data in dldTimeSteps and delayStage, select slices around to and also select in dldTimeSteps position around main core level peaks as well as around SBs we can get in some cases 5-10 energy points which are well-defined and they are intense enough to be used for find_bias_peaks routine. The only problem is that all of those will contain only one sample bias value, but this can be fed inside as an array of biases/energies.
This would be of help in case users forgot to do a proper energy calibration run (happening quite often :) )

kutnyakhov:
Three more tutorials for HEXTOF using new W(110) data sets from the public beamtime folder or also from Zenodo ([https://zenodo.org/records/12609441](https://zenodo.org/records/12609441)):
- Nº9 - an example of XPD workflow;
- Nº10 - an example of XPS workflow with a nice effect of BAM correction on t0;
- Nº11 - an example of energy calibration using SB to close #426;


rettigl:
Can you clean up all not required files, remove output from notebooks, and make sure tests pass? Also, is it necessary to have different versions of config files, or can they just be updated also for the old example?

rettigl (changes requested):
There are still some unused files (utils.py), and a lot of code in the Notebooks is rather redundant with other tutorials (energy calibration, saving, etc.), and not required for what you want to show. Also, it does not work in the current form outside maxwell (or even your account).

I restructured and streamlined the notebooks to some extend, see PR #524 
After we agree on a version, this should still be rebased and all older commits with notebook output removed, integrated into the docs, made sure they compile during docs building, etc.

kutnyakhov:
@rettigl thanks for cleaning up everything - your fixes branch now looks quite straightforward. Also, I like this comparison of both calibration methods, which at the end shows that the 4.5 eV range with SB calibration is even better than the 3 eV range with bias - it gives a proper 2.2 eV difference between W4f peaks.
utils.py file is not needed here - I just missed it during the last clean-up - but, in general, it should still be implemented at some point for spot size determination - I am preparing one more tutorial using different fluences, 2D Gauss fit and methods from Mac's paper for photoemission order.

In case we want to have also metadata - I can download a recent version from SciCat as json file that we can add to work outside of Maxwell.

rettigl:
> but, in general, it should still be implemented at some point for spot size determination - I am preparing one more tutorial using different fluences, 2D Gauss fit and methods from Mac's paper for photoemission order.

I suggest to implement such methods into the framework of the calibrator classes. I remember @steinnymir started working on such methods at some point. See https://github.com/OpenCOMPES/sed/issues/270 
The branches still exist and can be used as a starting point.


> In case we want to have also metadata - I can download a recent version from SciCat as json file that we can add to work outside of Maxwell.

We can add this later eventually if we want to, I don't think it that important for these tutorials, which aim at a specific functionality.

I will merge the fixes branch then here, rebase, and integrate these tutorials into the docs structure when I find time.

kutnyakhov:
> > but, in general, it should still be implemented at some point for spot size determination - I am preparing one more tutorial using different fluences, 2D Gauss fit and methods from Mac's paper for photoemission order.
> 
> I suggest to implement such methods into the framework of the calibrator classes. I remember @steinnymir started working on such methods at some point. See #270 The branches still exist and can be used as a starting point.
>
This utils.py comes from Steinn and I am/was just trying to finalize it :)
 
> > In case we want to have also metadata - I can download a recent version from SciCat as json file that we can add to work outside of Maxwell.
> 
> We can add this later eventually if we want to, I don't think it that important for these tutorials, which aim at a specific functionality.
>
Yes, I agree that this is not so important here.
> 
> I will merge the fixes branch then here, rebase, and integrate these tutorials into the docs structure when I find time.
>
Just a small correction before you merge to  No.11 notebook for SB calibration cell:
- I would use `sp_44498.ec.tof = res_corr['dldTimeSteps'].data` instead of `data.dldTimeSteps.data` - longer range - looks better :)
- we don't need `params = sp_44498.calibrate_energy_axis(' - you removed it and it is not needed any more

PR #523, opened 2024-11-19
Fix mpes incomplete files

### LINKED ISSUE #522: Skip bad h5 files in mpes loader
Automatically skip incomplete files rather than raising error

rettigl:
Adds exception handling for incomplete files that raise an IOError upon opening.

Closes #522

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -882,37 +882,37 @@ test = ["pandas[test]", "pre-commit", "pytest", "pytest-rerunfailures", "pytest-
 
 [[package]]
 name = "debugpy"
-version = "1.8.8"
+version = "1.8.9"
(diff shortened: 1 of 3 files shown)


zain-sohail (approved):
Instead of printing the error, I'd put this as logger warning or info. 
Secondly, I'd suggest using a more verbose error message, since you know the likely cause of this occurence: incomplete files.

rettigl:
> Instead of printing the error, I'd put this as logger warning or info. Secondly, I'd suggest using a more verbose error message, since you know the likely cause of this occurence: incomplete files.

The logging will happen once merged to the v1 branch (here we don't use the logging system yet). I can enhance the error message, thanks.

PR #525, opened 2024-11-22
Delay calibration fixes

### LINKED ISSUE #521: Allow adjustment of adc ranges without providing delay ranges
Currently, you have to provide delay_ranges to delay_calibration if you provide adc_ranges. This should be changed such that delay ranges are picked from a file if possible.

rettigl:
add option for either loading calibration data from file or from config
Closes #521

### DIFF
--- a/sed/core/processor.py
+++ b/sed/core/processor.py
@@ -1702,6 +1702,7 @@ class SedProcessor:
     def calibrate_delay_axis(
         self,
         delay_range: Tuple[float, float] = None,
+        read_delay_ranges: bool = True,
         datafile: str = None,
         preview: bool = False,
         verbose: bool = None,
@@ -1713,8 +1714,10 @@ class SedProcessor:
         Args:
             delay_range (Tuple[float, float], optional): The scanned delay range in
                 picoseconds. Defaults to None.
+            read_delay_ranges (bool, optional): Option whether to read the delay ranges from the
+                data. Defaults to True. If false, parameters in the config will be used.
             datafile (str, optional): The file from which to read the delay ranges.
-                Defaults to None.
+                Defaults to the first file of the dataset.
             preview (bool, optional): Option to preview the first elements of the data frame.
                 Defaults to False.
             verbose (bool, optional): Option to print out diagnostic information.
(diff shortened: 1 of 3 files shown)


rettigl:
I also removed the double pytest and coverage report in pull requests, as it prevents coverage reports in some cases apparently.

sed/core/processor.py
```diff
@@ -1702,6 +1702,7 @@ def align_dld_sectors(
     def calibrate_delay_axis(
         self,
         delay_range: Tuple[float, float] = None,
+        read_delay_ranges: bool = True,
```
zain-sohail, line 1705: If this flag is set to true by default, wouldn't it lead to an error for cases where data file is not available to read? Such as in the case of Flash data

rettigl: This case raises indeed an error. However, this function is anyways not intended for use with Flash data. It converts an ADC signal into a delay signal, and you directly use the delay stage values at Flash, no?
For using the stored calibration data regardless of the values in data files, I added this option, mostly for consistency with the config format. Honestly, I don't see a large use case for this option to load delay calibration data anyways, because typically they are different for each run, so any saved config is not relevant for a different scan.

zain-sohail: You're right about Flash data but this could pose a problem for future loaders utilizing ADC signal. If you find that this isn't a big issue, everything seems fine to me.

> For using the stored calibration data regardless of the values in data files, I added this option, mostly for consistency with the config format. Honestly, I don't see a large use case for this option to load delay calibration data anyways, because typically they are different for each run, so any saved config is not relevant for a different scan.

Does the user create different objects or calibration files for each run? Or is this an automated check based on metadata (such as runnumber)?

zain-sohail (approved):

sed/core/processor.py
rettigl: > Does the user create different objects or calibration files for each run? Or is this an automated check based on metadata (such as runnumber)?

Normally, the delay range is automatically read from each file, and applied with this function, without the need for saving the calibration dict. This uses the default adc_ranges set in the config. Just reconfiguring the adc_ranges and still taking delay_ranges from the file did not work if exectuted twice in the current main branch. This PR fixes this logic, so that this works as default. If one wants to instead use the saved config, the new flag has to be saved.

PR #532, opened 2024-12-17
Mpes process files from memory

rettigl:
Processing directly from the files for mpes is rather slow if reading from a filesystem with poor seeking performance (e.g. hdd, network drives). To overcome this, we load all files completely into memory for each binning step, and then access them from there.

### DIFF
--- a/sed/loader/mpes/loader.py
+++ b/sed/loader/mpes/loader.py
@@ -98,8 +98,6 @@ def hdf5_to_dataframe(
             search_pattern="Stream",
         )
 
-    test_proc.close()
-
     column_names = [alias_dict.get(group, group) for group in group_names]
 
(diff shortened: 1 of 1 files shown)


rettigl:
@zain-sohail, this is mpes-specific, maybe you can give it a brief look and commend/approve?

zain-sohail:
would this still work with smaller memory setups?

rettigl:
> would this still work with smaller memory setups?

We are talking about additional ~100 mbyte memory per compute core, compared to several GByte for the actual binning volume, so typically only an increase of 10% or less in memory consumption. You can always compensate with using less cores if memory is an issue.

zain-sohail (approved):
Tested. LGTM!

Issue #503, opened 2024-10-02
Dima's lab loader


zain-sohail:
Implement channel checking and skip that don't exist

zain-sohail:
New Metadata handling
Reading failing during ffill
schema check not working

PR #534, opened 2024-12-19
Hextof lab loader

zain-sohail:
This PR adds the lab loader requested in #503 . I tried to make minimal changes to the `FlashLoader` to make this work. The only major addition is the loader specific dataframe class and everything else stays approximately the same. So the lab data works with the flash loader but with`beamline` config as `cfel`.

An example config is provided to make this work. Since I took out some hardcoded paramters (was in TODO) into the config, I updated the config model slightly. 
Test data for this loading configuration still needs to be setup. I ask @kutnyakhov to provide a public file to perform this. Not sure if a tutorial is necessary or not.


src/sed/loader/flash/buffer_handler.py
```diff
@@ -124,13 +110,35 @@ def __init__(
...
+        core_beamline = config["core"].get("beamline")
+        self.DataFrameCreator: type[BaseDataFrameCreator] = None
+        if core_beamline == "pg2":
+            self.DataFrameCreator = DataFrameCreator
+        elif core_beamline == "cfel":
+            self.DataFrameCreator = CFELDataFrameCreator
+        else:
+            raise ValueError(f"Unsupported core beamline: {core_beamline}")
+
+    def _validate_h5_files(self, config, h5_paths: list[Path]) -> list[Path]:
```
zain-sohail, line 130: This validation was previously in BufferFilePaths and we had a discussion to move it from there. I find this location better (also was necessary due to restructure)

src/sed/loader/flash/utils.py
```diff
@@ -1,12 +1,6 @@
 from __future__ import annotations
 
 
-# TODO: move to config
-MULTI_INDEX = ["trainId", "pulseId", "electronId"]
-PULSE_ALIAS = MULTI_INDEX[1]
-FORMATS = ["per_electron", "per_pulse", "per_train"]
```
zain-sohail, line 7: these have now been moved to config/config model

src/sed/loader/flash/utils.py
```diff
@@ -39,32 +35,34 @@ def get_channels(
...
+                    f"Invalid format: {format_}. " f"Valid formats are: {formats_list + ['all']}",
+                )
 
         # Get the available channels excluding 'pulseId'.
         available_channels = list(channel_dict.keys())
         # pulse alias is an index and should not be included in the list of channels.
-        if PULSE_ALIAS in available_channels:
-            available_channels.remove(PULSE_ALIAS)
+        # Remove specified channels if they are present in available_channels.
+        channels_to_remove = ["pulseId", "numEvents"]
```
zain-sohail, line 62: Currently hardcoded. A better suggestion welcomed.

zain-sohail (commented):
Just few comments. Is this local metadata scheme also important for flash loader? Because then the code also needs to be updated there.

src/sed/core/config_model.py
```diff
@@ -26,6 +26,7 @@ class PathsModel(BaseModel):
 
     raw: DirectoryPath
     processed: Optional[Union[DirectoryPath, NewPath]] = None
+    meta: Optional[Union[DirectoryPath, NewPath]] = None
```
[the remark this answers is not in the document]
kutnyakhov: Fine for me. I just thought as it anyway would be one of the main folders inside the beamtime folder.

src/sed/loader/cfel/loader.py
```diff
@@ -403,7 +430,12 @@ def read_dataframe(
             filter_timed_by_electron=filter_timed_by_electron,
         )
 
-        self.metadata.update(self.parse_metadata(token) if collect_metadata else {})
+        if len(self.parse_scicat_metadata(token)) == 0:
+            print("No SciCat metadata available, checking local folder")
+            self.metadata.update(self.parse_local_metadata())
+        else:
+            print("Metadata taken from SciCat")
+            self.metadata.update(self.parse_scicat_metadata(token) if collect_metadata else {})
```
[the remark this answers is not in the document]
kutnyakhov: Fine for me. Just wanted to implement check if SciCat entries available then go for it, if not then check local folder to be compatible to older beamtimes.

PR #540, opened 2025-01-01
Move to uv

### LINKED ISSUE #531: Move packaging system to uv
uv (https://github.com/astral-sh/uv) is a new, extremely fast packaging manager. It also supports lock-files, caches in workflows etc. It would be worth checking if that can lead to performance improvements, and remove the complication of poetry.
https://stackoverflow.com/questions/79118841/how-to-migrate-from-poetry-to-uv-package-manager
https://pypi.org/project/poetry-to-uv/
https://www.loopwerk.io/articles/2024/migrate-poetry-to-uv/
https://github.com/astral-sh/uv/issues/5200

rettigl:
Moves the workflows and packageing system to uv.
Also uses dynamic versioning based on the git tags and commits, and removes the prerelease publishing

Closes #531


rettigl:
@zain-sohail release to test-pypi does not work right now, as it goes via your account with is refused because of missing 2FA. However, I believe the workflow is working in general and would be for main pypi. But maybe it's worth testing on test.pypi before merging. For this, you would have to create a "bublisher" there.

zain-sohail:
> @zain-sohail release to test-pypi does not work right now, as it goes via your account with is refused because of missing 2FA. However, I believe the workflow is working in general and would be for main pypi. But maybe it's worth testing on test.pypi before merging. For this, you would have to create a "bublisher" there.

Fixed that now. Currently uv is adding dev tags which pypi doesn't allow

```
Uploading sed_processor-0.4.1.dev342+gead1256-py3-none-any.whl
INFO     Response from https://test.pypi.org/legacy/:                           
         400 Bad Request                                                        
INFO     <html>                                                                 
          <head>                                                                
           <title>400 The use of local versions in                              
         <Version('0.4.1.dev342+gead1256')> is not allowed. See                 
         https://packaging.python.org/specifications/core-metadata for more     
         information.</title>                                                   
          </head>                                                               
          <body>                                                                
           <h1>400 The use of local versions in                                 
         <Version('0.4.1.dev342+gead1256')> is not allowed. See                 
         https://packaging.python.org/specifications/core-metadata for more     
         information.</h1>                                                      
           The server could not comply with the request since it is either      
         malformed or otherwise incorrect.<br/><br/>                            
         The use of local versions in                                           
         &lt;Version(&#x27;0.4.1.dev342+gead1256&#x27;)&gt; is not allowed. See 
         https://packaging.python.org/specifications/core-metadata for more     
         information.                                                           
                                                                                
                                                                                
          </body>                                                               
         </html>                                                                
ERROR    HTTPError: 400 Bad Request from https://test.pypi.org/legacy/          
         Bad Request                                                            
```
This is probably not critical to change unless the same thing happens on main. I've not looked at uv in detail yet.

.github/workflows/benchmark.yml
```diff
@@ -16,31 +19,32 @@ jobs:
...
 
-      - name: Install project dependencies
-        run: poetry install
+      - name: Install dependencies
+        run: |
+          curl -LsSf https://astral.sh/uv/install.sh | sh
 
-      - name: Install project dependencies
-        run: poetry install
+      - name: Install package
```
zain-sohail, line 40: I saw that there's already an official action by uv
`uses: astral-sh/setup-uv@v5`

zain-sohail: https://docs.astral.sh/uv/guides/integration/github/#installation

rettigl:
> Fixed that now. Currently uv is adding dev tags which pypi doesn't allow

I see. For the actual releases, this won't be a problem, but the workflow dispatch action does not make sense then. I actually tested the same approach for the specsanalyzer package already, and it works on release, so we should be good. I will update the workflow accordingly.

zain-sohail (approved):
LGTM!
Any changes, we can also do later. I think the maintainer docs need to be updated.

.github/workflows/benchmark.yml
rettigl: I tested this, but did not see any advantage. On the contrary, it tries to use some kind of cache, this however takes longer than without, and also gets confused during the matrix tests. I'll revert to the previous situation.

zain-sohail: Matrix method requires (from the link)

```
      - name: Install uv and set the python version
        uses: astral-sh/setup-uv@v5
        with:
          python-version: ${{ matrix.python-version }}
```

Cache should generally be faster after first time, no?

rettigl:
> LGTM! Any changes, we can also do later. I think the maintainer docs need to be updated.

I also updated the docs now. If you want, you can give it a final look.

.github/workflows/benchmark.yml
rettigl: The caches it created are only 8mbyte or so, so they don't contain anything important. UV hard-links the packages to versions available already on the system, thus the installation is super-fast, the whole setup takes less then 10s. Using caches, I am not sure we will be using the most-recent package version all the time, thus I prefer the current workflow. I run the multiversion tests now every week to see that thinks keep working if packages update. This way, we don't have to constantly update a lockfile (which you can also use with uv, but I don't see the real advantage for our purposes).

zain-sohail:
> > LGTM! Any changes, we can also do later. I think the maintainer docs need to be updated.
> 
> I also updated the docs now. If you want, you can give it a final look.

Everything looks ok. I just didn't understand why the prereleases were removed from release workflow. Was there a problem with them?

rettigl:
> I just didn't understand why the prereleases were removed from release workflow. Was there a problem with them?

I did not really see the advantage of having them. If anyone wants a version more recent than the last release, you can always install from git. The problem I had several times with the old workflow was that if you merge a second PR while the pre-release workflow was still running after merging one PR, the version bump would not work because of the intermediate merge commit. As I don't see a reason for having the pre-releases, easiest solution was to remove them. Also, the dynamic dev versions apparently are not allowed on pypi.

rettigl:
Here https://discuss.python.org/t/generating-pep-440-compliant-development-releases-containing-githash/12302/4 it is stated:

> The recommendation is for development releases to not be distributed via a package index, so Pip won’t have to resolve a development version. I recommend serving the packages using a simple HTTP server (internal or with basic auth), so Pip’s resolution can be bypassed.

zain-sohail:
My reason was simply that many users are struggling with git but doing pip install is generally known to most. However, as you pointed out, this is not the recommendation.
We can merge this.

PR #541, opened 2025-01-06
remove empty pulses from timed dataframe, and bring back old behavior

rettigl:
This brings back more or less the old behavior of the timed dataframe, containing only pulses with electrons.
This empirically provides better normalization.


rettigl:
@zain-sohail This is a rather dirty fix, and does not provide the option to maintain the current behavior. In principle, one could also remove the saving of timed dataframe parquets with this again. Feel free to add to it...

rettigl:
@zain-sohail can you have a look at this? I would like to finalize v1 rather soon, and I think this is the only blocker left.

zain-sohail:
The other solution should still exist. I'll update this branch to include a flag for wanting such forms of timed dataframes for when special users want to use it.

rettigl (commented):
@zain-sohail LGTM, maybe add a few lines to the docu about this at the respective place.
Feel free to merge.

zain-sohail (approved):

PR #549, opened 2025-01-13
Pydantic error handling

### LINKED ISSUE #544: Throw a better error pydantic
Not every user knows what pydantic is or does.

Currently, if the config is incorrect, pydantic just throws multiple error messages which are extremely verbose. The simplest solution is to catch and first log for the user that their config is at fault and not the code.
Second idea would be to combine these validation errors and throw something more understandable to the user.

Issue brought up by dima.

zain-sohail:
closes #544

### DIFF
--- a/src/sed/core/config.py
+++ b/src/sed/core/config.py
@@ -24,17 +24,6 @@ USER_CONFIG_PATH = user_config_path(appname="sed", appauthor="OpenCOMPES", ensur
 logger = setup_logging("config")
 
 
-class ConfigError(Exception):
-    """Exception raised for errors in the config file."""
-
-    def __init__(self, message: str):
-        self.message = message
-        super().__init__(self.message)
-
-    def __str__(self):
-        return self.message
-
-
 def parse_config(
     config: dict | str = None,
     folder_config: dict | str = None,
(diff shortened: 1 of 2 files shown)


src/sed/core/config.py
```diff
@@ -146,9 +159,19 @@ def parse_config(
...
+        config_model = ConfigModel(**config_dict)
+        return config_model.model_dump(exclude_unset=True, exclude_none=True)
+    except ValidationError as e:
+        error_msg = (
+            "Invalid configuration file detected. The following validation errors were found:\n"
+        )
+        for error in e.errors():
+            error_msg += f"\n- {' -> '.join(str(loc) for loc in error['loc'])}: {error['msg']}"
+        logger.error(error_msg)
+        raise ConfigError(error_msg) from e
```
rettigl, line 174: Isn't it sufficient and actually better to just log the error here and raise the original error? Why creating a new error type?

tests/test_config.py
rettigl, line 1: This does not seem to work.

rettigl (commented):
I believe just adding a logger mesage is sufficient, rather than creating a new error type.
You can also use a ValueError with from e

src/sed/core/config.py
zain-sohail: That's a good point. I just found that it makes thing very explicit but it can be removed.

zain-sohail: I have switched to a ValueError from ConfigError as suggested, as we are not trying to programatically catch the ConfigError so it's redundant.
It's possible to just log and not throw any error but then the testing has to be updated further.

PR #553, opened 2025-01-17
update tests for env variables

### LINKED ISSUE #552: env var tests delete .env files
by running the test for the env var system, the user .env file gets deleted.
@zain-sohail can you fix this?

zain-sohail:
closes #552

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -15,7 +15,6 @@ from sed.core.config import parse_config
 from sed.core.config import read_env_var
 from sed.core.config import save_config
 from sed.core.config import save_env_var
-from sed.core.config import USER_CONFIG_PATH
 
 test_dir = os.path.dirname(__file__)
 test_config_dir = Path(f"{test_dir}/data/loader/")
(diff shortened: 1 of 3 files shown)


tests/test_config.py
```diff
@@ -336,14 +328,20 @@ def test_env_var_save_and_load(cleanup_env_files):  # noqa: ARG001
     assert read_env_var("OTHER_VAR") == "other_value"
 
 
-def test_env_var_not_found(cleanup_env_files):  # noqa: ARG001
+def test_env_var_not_found(mock_env_file):  # noqa: ARG001
     """Test behavior when environment variable is not found"""
     assert read_env_var("NONEXISTENT_VAR") is None
```
rettigl, line 333: This tests fails if I do "export NONEXISTENT_VAR=test" before, but this is a rather artificial case...

rettigl (approved):
I don't understand all details, but it fixes the problem, and tests still pass, so LGTM

tests/test_config.py
zain-sohail: Yes this won't be the case at least in the github workers, and if it happens, we can adjust.

PR #554, opened 2025-01-19
Fix release and update documentation

rettigl:
- Update GitHub Actions workflow for documentation: Don't set pre-releases as stable
- Modify pyproject.toml for release adjustments


rettigl:
@zain-sohail With this docs built for pre-releases (aN suffix) will not be tagged as "stable". Also, python versions will be correctly tagged now with the new build system

zain-sohail (approved):
Seems that the action works. Any reason why you merge to v1?

rettigl:
> Any reason why you merge to v1?

The changes to pyproject would require additional merges probably, and we want to merge v1 anyways...

PR #555, opened 2025-01-20
use user platformdir also for user config

rettigl:
change user and system config files to config_v1.yaml add system wide .env


zain-sohail (approved):
Everything looks good. Prob an additional test for system config path but it's not necessary

rettigl:
> additional test for system config path but it's not necessary

Added

PR #557, opened 2025-01-26
Fix normed dtype

rettigl:
Sets the dtype of normalized data to that of unnormalized data.
Currently, it gets the dtype of the normalization histogram

### DIFF
--- a/src/sed/binning/binning.py
+++ b/src/sed/binning/binning.py
@@ -479,13 +479,7 @@ def normalization_histogram_from_timed_dataframe(
     Returns:
         xr.DataArray: Calculated normalization histogram.
     """
-    bins = df[axis].map_partitions(
-        pd.cut,
-        bins=bin_centers_to_bin_edges(bin_centers),
-    )
-
-    histogram = df[axis].groupby([bins]).count().compute().values * time_unit
-    # histogram = bin_dataframe(df, axes=[axis], bins=[bin_centers]) * time_unit
+    histogram = bin_dataframe(df, axes=[axis], bins=[bin_centers]) * time_unit
 
     data_array = xr.DataArray(
         data=histogram,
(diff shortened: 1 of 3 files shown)


src/sed/core/processor.py
```diff
@@ -2375,13 +2377,25 @@ def get_normalization_histogram(
...
+                    threads_per_worker=self.config["binning"]["threads_per_worker"],
+                    threadpool_api=self.config["binning"]["threadpool_API"],
                 )
             else:
                 self._normalization_histogram = normalization_histogram_from_timed_dataframe(
                     self._timed_dataframe,
                     axis,
                     self._binned.coords[axis].values,
                     self._config["dataframe"]["timed_dataframe_unit_time"],
+                    hist_mode=self.config["binning"]["hist_mode"],
```
zain-sohail, line 2393: This seems repeated. Probably can go out of the loop

src/sed/binning/binning.py
```diff
@@ -475,17 +476,12 @@ def normalization_histogram_from_timed_dataframe(
             histogram.
         bin_centers (np.ndarray): Bin centers used for binning of the axis.
         time_unit (float): Time unit the data frame entries are based on.
+        **kwds: Additional keyword arguments passed to the bin_dataframe function.
 
     Returns:
         xr.DataArray: Calculated normalization histogram.
     """
-    bins = df[axis].map_partitions(
```
zain-sohail, line 482: is this removed due to the updated dask version?

rettigl: I am using our optimized binning now for the timed dataframe. This is somewhat faster, does the sequential binning using the num_cores parameter, and shows the progress bar. The previous solution used the pandas cut to define bins, which requires bin edges rather than bin centers as our function.
I once checked that they produce very similar results (a very tiny difference was there, I think, because of different inclusion/exclusion of the bin edges into either left or right bin).

src/sed/core/processor.py
rettigl: I changed the structure now to repeat less code

zain-sohail (approved):

PR #562, opened 2025-02-25
Fix dataset tests

zain-sohail:


rettigl:
It fails on python versions 3.11, 3.12, with errors in creating temp directories. Not sure about the reason.

rettigl:
There are strong interconnections and race conditions between the different tests, as they all access the same datasets.json files as defined by the static dm. This needs also to be resolved, parallel running test very often fail randomly.

PR #563, opened 2025-03-02
Update widths of figures

zain-sohail:


rettigl (changes requested):
Figures appear a bit odd-shaped at places, in particular this one: 
image
I suggest to keep similar aspect ratios as before, and just scale down.

tutorial/9_hextof_workflow_trXPD.ipynb
rettigl, line 1: Changel L416 as well

rettigl (approved):
Should be fixed now.

PR #564, opened 2025-03-03
Add len and count rate methods to flash loader

zain-sohail:


zain-sohail:
- [] add sp.info to show the metadata

Aserhisham (approved):

PR #567, opened 2025-03-24
Update pynxtools

rettigl:
Update Nexus config to newest release

PR #568, opened 2025-03-25
Speedup mpes timed dataframes

rettigl:
replace for loop with numpy indexing


rettigl:
@zain-sohail test and linting failures are covered in #567

rettigl:
@zain-sohail This is only a small fix, so review should not take long

PR #571, opened 2025-04-06
add logo to README and docs

zain-sohail:


docs/_static/logo.png
rettigl, line 1: 1 MByte for a logo seems exhaustive to me. I suggest to downsample to ~50-100 kByte (then force push to remove the large file).

README.md
```diff
@@ -1,3 +1,7 @@
+<h1 align="center">
+<img src="docs/_static/logo.png" width="200">
+</h1><br>
```
rettigl, line 3: This big logo looks a bit out of place here, I find. I also don't find logos in readme files regularly. Mabye we put such a larger logo into the docs entry page instead?

docs/conf.py
```diff
@@ -13,7 +13,7 @@
 
 import tomlkit
 
-from sed import __version__
+from sed import __version__  # type: ignore[attr-defined]
```
rettigl, line 16: Why this?

docs/conf.py
```diff
@@ -105,6 +105,9 @@ def _get_project_meta():
 #
 html_theme = "pydata_sphinx_theme"
 
+html_logo = "_static/logo.png"
+html_favicon = "_static/favicon.ico"
```
rettigl, line 109: Not sure where and how this is used (?)

rettigl (commented):
We can also add the logo as a social medial preview to the repo config.

docs/conf.py
zain-sohail: It should appear on the docs top left and the favicon is the icon that shows on website tabs.

docs/conf.py
zain-sohail: mypy complains otherwise. It probably didn't before because this was a committed change from before

README.md
zain-sohail: We can put a smaller one. I saw one on numpy github so thought to put it. Easy to copy for people when necessary this way too.

docs/conf.py
zain-sohail: image

README.md
rettigl: I think this is good now.

docs/conf.py
rettigl: You are right, now it shows. Last time I did not see it, maybe browser cache or so.

docs/conf.py
rettigl: This is not regularly checked by our pipelines. I still don't understand why there is a problem, but fine.

rettigl (approved):
LGTM

PR #573, opened 2025-04-14
add python 3.13

rettigl:


pyproject.toml
```diff
@@ -40,7 +41,7 @@ dependencies = [
     "matplotlib>=3.5.1",
     "natsort>=8.1.0",
     "numba>=0.55.1",
-    "numpy>=1.18,<2.0",
+    "numpy>=1.18",
```
zain-sohail, line 44: There's some library here that doesn't work with numpy 2.0 right?

zain-sohail (approved):
Tests are passing so LGTM

pyproject.toml
rettigl: I ran the tests with numpy>2, and everything worked, so I believe this was fixed. Thus I removed the restriction.

PR #574, opened 2025-05-06
update pynxtools version and adjust nexus config

rettigl:
Small updates to the nexus definitions, now very close to being officially standardized


zain-sohail (approved):
LGTM since it's definition updates already agreed upon by the community.

PR #575, opened 2025-05-06
Fix pose adjustment

rettigl:
Fix for the application of pose corrections without using a transformation. Also added a test for it.

### DIFF
--- a/src/sed/calibrator/momentum.py
+++ b/src/sed/calibrator/momentum.py
@@ -874,6 +874,10 @@ class MomentumCorrector:
             stackaxis=0,
         ).astype("float64")
 
+        # reset image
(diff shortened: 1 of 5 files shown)


rettigl:
@zain-sohail This fixes application of pose adjustments without distortion correction. Before, the correction was always loaded from the backup if it was calculated before. Now, it can be properly disabled with use_correction=False.

zain-sohail:
> @zain-sohail This fixes application of pose adjustments without distortion correction. Before, the correction was always loaded from the backup if it was calculated before. Now, it can be properly disabled with use_correction=False.

zain-sohail (approved):
That was definitely a hidden bug. Thanks for adding tests to check intended behavior.

PR #576, opened 2025-05-06
Calculate temporal resolution

rettigl:
Estimate temporal resolution from the pulse durations from elabFTW metadata if available


rettigl:
@zain-sohail this is a small change for mpes metadata, if you can quickly give it your ok.

zain-sohail (approved):
Logic seems fine. The method in this module seem to span a lot so it's hard to read so it'd be good to have a couple of helper methods. But it's not necessary.
I also noticed there are no test cases for the current update (and others). If it's not a key element, then those aren't quite necessary.

rettigl:
I know there are no strict tests for this, but it's also very specific to our lab setup, and also depend on the ELN setup.

